PR #119, opened 2015-01-19
solver docs

scopatz:
associated with cyclus/cyclus#1070


gonuke:
I don't think these docs match the version that was ultimately merged in cyclus/cyclus#1070.  (or am I reading that wrong?)

PR #141, opened 2015-05-13
Make archetype docs even prettier

rwcarlsen:
* Use subheadings for State variables, and schema sections rather than just
  bold text on its own line.

* Move misc annotations into their own "Other Info" subheading section rather
  than just below the doc paragraph.  This section is placed below everything
  except the schema

* Move state variable units annotation onto the same line as state var header
  right next to the variable type - e.g. :[statevarname]: ``double`` kg.

* Don't print the tooltip if there is a doc annotation.  tooltip is a fallback
  only if necessary.  This eliminates redundant/annoying info and makes the docs
  much nicer to look at.

### DIFF
--- a/source/cyclusagent.py
+++ b/source/cyclusagent.py
@@ -261,18 +261,24 @@ class CyclusAgent(Directive):
         if 'doc' in self.annotations:
             self.lines += self.annotations['doc'].splitlines()
         self.lines.append('')
+
+    def append_otherinfo(self):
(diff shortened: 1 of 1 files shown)


mbmcgarry:
This looks great!  Thanks @rwcarlsen !

mbmcgarry:
manually merged into source branch with commit 2041723..530d685

PR #147, opened 2015-05-18
updating 'other info' for cleaner key names

gidden:
You now get better looking "other info" listings

tmp2


gidden:
ping

mbmcgarry:
So this looks nice, but in the implementation there are now two sets of xml snippets that are not the same:  the state var info and the xml input schema.  Is there value to keeping them both here? I think a user who is inexperienced with xml would be confused as to which one to emulate.

PR #151, opened 2015-05-27
First attempt at page linking to projects using Cyculs

gonuke:
This probably isn't quite ready.  Minimally we need to confirm titles for all the projects, and then we can push without links. 

Ultimately, it would be good to include the following for each project:
* logo
* correct title with link
* 1-3 sentence description

These can be PR'd by the project owners.

### DIFF
--- a/source/projects/index.rst
+++ b/source/projects/index.rst
@@ -1,0 +1,13 @@
+|Cyclus|-Related Projects
+==========================================
+
+This is a partial list of projects that are known to be using |Cyclus| as part
+of their research.
+
+* Consortium for Verification Technology (U. Wisconsin-Madison with U. Michigan)
+* Bright-lite (U. Texas-Austin)
+* ???? (U. Tennessee)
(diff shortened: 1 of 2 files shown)


source/projects/index.rst
```diff
@@ -0,0 +1,13 @@
...
+==========================================
+
+This is a partial list of projects that are known to be using |Cyclus| as part
+of their research.
+
+* Consortium for Verification Technology (U. Wisconsin-Madison with U. Michigan)
+* Bright-lite (U. Texas-Austin)
+* ???? (U. Tennessee)
+* Fuel Cycle Options Transition Analysis (LLNL, U. California-Berkeley, U. Wisconsin)
+* Fuel Cycle Optimization (U. Wisconain-Madison)
```
opotowsky: @gonuke for a simple typo like this, is it a best practice to ask you to correct it?

gonuke: yep.  and I will

gonuke: Although for this PR, it's a place holder that is inviting others to add/correct things

PR #157, opened 2015-06-08
Revert "remove name and id columns completely"

scopatz:
This reverts commit 69a071e0490300756e64b13f550cb593ea84959c.  

This fixes should fix the error with cymetric building caused by removing these columns.  These columns where put there originally because they include information that is not present in the other columns.  Namely, the id is the `cyclus::DbTypes` enum value and the name is the name in the namespace.  It is not possible to 100% infer these from the other columns - which is why they were included.  This reverting prevents there from being more than one table of this information anywhere in the ecosystem.


gonuke:
We discussed this in person, so let me capture it here.
* A single authoritative place for this information is definitely the right choice
* Cyclus is probably the best place as it binds it most closely to the place where the underlying information is changing, and both cymetric and cyclus.github.com depend on it already
* a cymetric dependency on cyclus.github.com is non-intuitive
* ideally, the version on cyclus.github.com (or at least one version) is better without all the extra detail - a typical archetype developer doesn't need to understand the additional detail and may in fact be confused by that detail
* it would be great if this could be auto-generated from the place where these are defined in code

Given this ultimate goal, we determined that a reasonable next step was to move this to cymetric.  

It was short-sighted to not recognize the way that it broke prior versions of Cymetric., which we should address, even if it temporarily results in duplicated data.

scopatz:
Hi @gonuke, I agree that the current dependency hierarchy is less than ideal (though not wrong, per say) and that the best way would be to generate from the code - if possible (though I don't think it is completely without embedding data in comments or adding some refection somewhere). However, I disagree with the results of the in person though not all of the points for a couple of reasons.  

> ideally, the version on cyclus.github.com (or at least one version) is better without all the extra detail - a typical archetype developer doesn't need to understand the additional detail and may in fact be confused by that detail

This was the only place in the documentation that an archetype developer could find the name of a DbType. Maybe not every archetype dev needs to create custom tables or cast things back and forth, but some do and so I think that this information needs to be present in the docs somewhere for completeness. An archetype developer should not have to go through digging through header files and make guesses about what the names are.  

I also don't think that duplicating data is a good idea.  Once it is done, the barrier to remove it is much higher than duplicating in the first place. It also leaves open an area where cymetric can be behind or broken. In the long run, this could give cymetric a bad reputation for not compiling or being up to date.

If duplication is chosen, then the duplicating process really needs to become part of CEP3 and the whole release structure.  Maybe I missed them but I didn't see any PRs to that effect.

Personally, I think that the quickest and most robust solution is just to leave it as it was until such a time that a clean and robust solution can be implemented.

In any event, this is completely blocking all of my work right now.

rwcarlsen:
Closed in favor of #162

PR #158, opened 2015-06-09
dev tutorial info

gidden:


mbmcgarry:
Thanks @gidden !
Note to all: It looks like the Users guide is still awaiting instructions.

PR #159, opened 2015-06-10
Ans 2015 tut

gonuke:
A complete set of files for the ANS2015 User tutorial - exercise #1!

May need spell checking? Probably in realtime?


gidden:
looks good! switching partitions to publish.

gidden:
(note, I did not spell check -- I agree that we can fix that on the fly)

gidden:
and we're live!

PR #163, opened 2015-06-16
noticed there are other places to bump versions

gidden:


opotowsky:
Good catch! I noticed cymetric's version is still 0.0. Merging, but I won't publish since cymetric's master branch is broken. Thanks @gidden!

PR #168, opened 2015-08-03
adding a docker file, it's currently broken

gidden:
Cymetric is still broken, but I think there's something more here (I get a DivisionByZero error). Could be some Python 3 issues too..

Here's a [gist](https://gist.github.com/gidden/2329b706ec716511b339)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+
+FROM cyclus/cycamore
+
+RUN apt-get -y --force-yes update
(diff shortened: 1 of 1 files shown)


rwcarlsen:
A good start, but going a bit further - I was thinking of something more along the lines of #169.  We do need to figure out cymetric issues first :\

gidden:
#169 is fine, closing

PR #169, opened 2015-08-03
docker build option

rwcarlsen:
This adds a dockerfile that is based on cyclus/cycamore that adds other deps necessary for a site build.  The image from this dockerfile will need to be pushed up to docker hub when we get cymetric fixed.  This adds 2 targets html-docker and gh-preview-docker to the makefile that builds the site inside the fuelcycle.org-deps image and leaves the built site files in the host computer's working directory exactly as if they had run the make html command like normal without docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,9 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 
+html-docker gh-preview-docker:
+	docker run -w /cyclus.github.com -v $(PWD):/cyclus.github.com cyclus/fuelcycle.org-deps make gh-preview
+
 gh-clean gh-revert clean:
 	-rm -rf $(BUILDDIR)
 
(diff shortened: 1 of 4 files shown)


gidden:
I get the following (pre-cymetric build) errors:
```
Step 3 : WORKDIR /
 ---> Using cache
 ---> 310adad546a8
Step 4 : RUN git clone https://github.com/cyclus/cymetric
 ---> Running in eb6b6abe8867
2015/08/03 16:16:21 /var/lib/docker/aufs/mnt/eb6b6abe886764e1897c0ac2e95554106e5e4bc029af86f9e1f854b1c55261be is not within /var/lib/docker/aufs/mnt/eb6b6abe886764e1897c0ac2e95554106e5e4bc029af86f9e1f854b1c55261be
```

gidden:
@rwcarlsen, `make docker-html` now works on my machine on the upstream/docker branch. want to pr that one instead?

rwcarlsen:
Yes

PR #172, opened 2015-08-17
Add QR code for fuelcycle.org

gonuke:
We don't need to use this on the site (yet?) but it's probably the best place for safekeeping.


rwcarlsen:
cherry-picked as 54c7047.

PR #176, opened 2015-09-07
Change status of cep25

gonuke:


gonuke:
Thanks @gidden - sorry this took so long

PR #179, opened 2015-10-06
update docs for cyclus/cyclus#1193

gidden:


rwcarlsen:
cherry-picked as c74e440

PR #183, opened 2015-12-10
removing install instructions from tutorial (to be placed in cymetric…

opotowsky:
… repo)

companion: https://github.com/cyclus/cymetric/pull/94


rwcarlsen:
cherry-picked as fbd39de

PR #185, opened 2016-01-22
created section for third-party archetypes, and added mbmore:RandomEn…

mbmcgarry:
…rich and mbmore:RandomSink to the list

### DIFF
--- a/source/user/index.rst
+++ b/source/user/index.rst
@@ -65,7 +65,8 @@ Archetypes
 Collections of :term:`archetypes <archetype>` are provided by various teams
 and individuals for use by |Cyclus| users.
 
-**Cycamore**
+Cycamore
+~~~~~~~~
    The Cycamore library of archetypes is provided by the |Cyclus| development
    team as a basic set of archetypes for modeling simple nuclear fuel cycles.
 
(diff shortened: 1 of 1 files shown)


source/user/index.rst
```diff
@@ -74,6 +75,19 @@ and individuals for use by |Cyclus| users.
...
+
+.. toctree::
+   :maxdepth: 2
+
+   cyclusagents
+
+
+Third-Party Archetypes
+~~~~~~~~~~~~~~~~~~~~~~
 **Bright-lite**
```
opotowsky: @mbmcgarry We discussed on the dev call that we won't advertise Bright-lite until it's refactored. Maybe we can just comment this out for now?

opotowsky: (Whoops, even though it was already on here)

mbmcgarry: If we want to comment out Bright-lite and Cyborg for now then I can do that, but they were already listed on the website prior to this PR, I just reorganized slightly.

rwcarlsen:
pinging the original reviewer - @opotowsky...

opotowsky:
Sorry I forgot to come back to this, @mbmcgarry - in now!

PR #189, opened 2016-04-06
fix tab and incorrect python3 string handling

rwcarlsen:
The website build was broken by some additions for json snippet generation/printing in #186.  This should fix those issues.

### DIFF
--- a/source/cyclusagent.py
+++ b/source/cyclusagent.py
@@ -256,10 +256,11 @@ def build_json_sample(cpptype, schematype=None, uitype=None, names=None, units=N
             name = names
         d_type = _type(t, schematype or uitype)
         d_type = uitype if uitype in special_uitypes else d_type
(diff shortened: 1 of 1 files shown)


rwcarlsen:
@scopatz - will this change break python2 - I tried to do a preliminary check that seemed fine, but perhaps a full python2 system would provide more assurance.

Baaaaam:
@rwcarlsen I confirme this is working using:
Python 3.5.1
Sphinx v1.4
Cloud Version: 1.7.1

don't know if we want to wait python2 compatibility validation....

Baaaaam:
working with : 
Python 2.7.10
Sphinx v1.4
Cloud Version: 1.7.1

merging

scopatz:
Yeah, I couldn't see a way this would break things @rwcarlsen

PR #190, opened 2016-04-07
Add details for improving docker usage experience:

rwcarlsen:
* Add more instructions to readme for updating website deps image

* Modify/add "serve" make targets to support python3 in addition to allowing
  servince pre-built files (e.g. to enable serving a docker build)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ help:
 	@echo "  gh-publish-only   to push from source branch to master branch, assuming already built"
 	@echo "  docker-html       to use docker to build HTML in directory $BUILDDIR for testing"
 	@echo "  docker-gh-publish to use docker to build final version and push from source branch to master branch"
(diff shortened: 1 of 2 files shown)


source/cyclusagent.py
```diff
@@ -474,7 +474,7 @@ def append_statevars(self):
                 self.lines.append(ind + ':{0}: {1}'.format(key, val))
             self.lines.append('')
 
-            self.lines += [ind + '**XML:**', '', ind + '.. code-block:: xml', '']
+            self.lines += ['**XML:**', '', ind + '.. code-block:: xml', '']
```
gonuke: Is this just an aesthetic change, or does removing the indent change parsing/processing behavior?

rwcarlsen: Oops - just experimenting that I didn't mean to commit - reverted.

opotowsky:
Thanks @rwcarlsen !

PR #191, opened 2016-04-14
add rwc archetypes lib description

rwcarlsen:


source/user/index.rst
```diff
@@ -117,6 +117,47 @@ Third-Party Archetypes
       or at Random timesteps, and quantity requested can be varied using a
       Gaussian distribution function.
  
+**rwc**
```
mbmcgarry: perhaps heading title could be in caps just for asthetics (ie: RWC)

source/user/index.rst
```diff
@@ -117,6 +117,47 @@ Third-Party Archetypes
...
+
+    `:rwc:LookInst
+    <https://github.com/rwcarlsen/rwc-archetypes/blob/master/look_inst.h>`_
+    This archetype allows the user to specify an arbitrary power capacity time
+    series that the institution will automatically deploy power generating
+    facilities (of multiple types) to match. It uses a look-ahead mechanism to
+    see if its chosen deployments were not "good" and adjusts its deployment
+    decisions accordingly. This archetype currently requires
+    rwcarlsen/cyclus#restart branch of Cyclus to work.
+
```
mbmcgarry: I think it looks good with a whitespace between each archetype, but this is inconsistent with the mbmore descriptions above.  Any chance you want to throw a whitespace in the mbmore examples to make it all consistent?

mbmcgarry: Also inconsistent between the sections is bullet-ing for each listed archetype

mbmcgarry:
Can you reorder the archetypes with some logic - perhaps alphabetical ?

mbmcgarry:
General question ( @rwcarlsen @scopatz @gonuke @opotowsky @Baaaaam ): As we add 3rd party archetypes, do we want to require any quality control (unit tests, code review, formatting etc?)

opotowsky:
The general rule is it's up to the user to defend their work if they use certain archetypes, but I can see why we would want to have a standard for what we feature on the website.

scopatz:
I agree with @opotowsky - 3rd party libs are by definition outside of the cyclus org's qc.

rwcarlsen:
I think we don't worry about it - especially at these earlier stages in
archetype community growth.  When the list becomes sufficiently large,
perhaps we can become more selective.  Maybe it would be worth throwing in
a one-sentence disclaimer at the top of the "third-party" archetypes
section though.

rwcarlsen:
I added a disclaimer to the 3rd party section and cleaned up formatting in @mbmcgarry and rwc archetypes sub-sections.  @mbmcgarry - the order is explicit - it was chosen in order of awesomeness - or in the order I would want a user to read.  But I don't feel super strongly about it - I'd be happy to reorder it as long as the institution is first.

mbmcgarry:
@rwcarlsen while I do appreciate the urge to order by awesomeness, I imagine as more archetypes are added it may become difficult to appropriately rank them by awesome. I'd prefer something along the lines of alphabetical within Region->Insititution->Facility or similar

rwcarlsen:
Fair enough - re-ordered.

mbmcgarry:
looks great, and thanks for standardizing the formatting for the entire Third-Party Archetypes section!

PR #208, opened 2016-10-14
Including using start/end flags

Baaaaam:
change include by line number to include by flags


scopatz:
Seems reasonable.

scopatz:
Thanks!

PR #211, opened 2016-10-17
fix flag name grab

Baaaaam:
correct flag name


Baaaaam:
closing as it is redundant with #214

PR #215, opened 2016-10-21
CEP26

scopatz:
See diff. Also by @FlanFlanagan


scopatz:
Hello All, Since this is only a draft, are there any objections to merging this for historical purposes?

scopatz:
Note that this is not accepted, it is just being listed as proposed.

PR #227, opened 2017-01-09
WIP: update CEP3 to actual release process.

bam241:
Todo list:
- [x] conda-forge update procedure
- [x] remove database type (it is now automatic)
- [x] add PyNe and Oxygen decencies for the maintenance task
- [x] update Cyclus/Cycamore version file to src/version.h.in and src/cycamore_version.h.in
- [x] add Dory update nd worker relaunch procedure 
- [x] add instruction commit with ABI/API acceptable change ?! (as we did for 1.5.0) ?

### DIFF
--- a/source/cep/cep3.rst
+++ b/source/cep/cep3.rst
@@ -139,6 +139,26 @@ Release Process
       $ export CYCAMORE_DIR=/path/to/cycamore
       $ ./maintenence.sh -r -v X.X.X # X.X.X is *this* version
 
+    .. note:: 
+
+          If mainteance script fails because of an ABI failure that are cause by
+          a compiler update, you might want to accept them and procceed with the
+          release with those. to do so you need to generate the new symbols and
+          commit them:
+          
+          #. First make sure those change can be ignore by email the dev-list
+          
+          #. if the dev-list agreed to those changes, update the symbol and
+             commit the new one:
+
+          .. code-block:: bash
+            
+                $ cd $CYCLUS_DIR/release
+                $ ./smbchk.py --update -t X.X.X # X.X.X is *this* version
+                $ git add symbols.json
+                $ git commit -m "Describe here all the change on the ABI"
+
+
 #. Commit all changes for all projects.
 
     .. code-block:: bash
--- a/source/cep/cep3.rst
+++ b/source/cep/cep3.rst
@@ -222,7 +222,22 @@ Release Process
 
 #. Commit all changes to ``cyclus.github.com`` and ``make gh-publish`` 
 
-#. Update Dory Cyclus/Cycamore version and relaunch Dory worker.
+#. Update Dory Cyclus/Cycamore version and relaunch Dory worker. To do this you
+   need a acces to the dory server, the go toolchain as well as cde installed on
+   your computer.
+    
+   .. code-block:: bash
+      $ git clone https://github.com/rwcarlsen/cloudlus.git
+      $ cd cloudlus
+      $ go install ./cmd/cloudlus
+      $ cd misc/fuelcycle.org
+      $ make
+      $ ssh dory.fuelcycle.org ‘mv cyc-cde.tar.gz cyc-cde.tar.gz_bkp’
+      $ scp cyc-cde.tar.gz dory:fuelcycle.org:./
+      $ ssh dory.fuelcycle.org
+      $ ps -fe | grep cloudlus | grep work | grep ':80' | cut -d" " -f6 | xargs kill -9
+      $ rm -rf worker-*
+      $ ./launch.sh 2
 
 #. Send out an email to `cyclus-dev` and `cyclus-users` to announce the release!
 
--- a/source/cep/cep3.rst
+++ b/source/cep/cep3.rst
@@ -266,6 +266,15 @@ Release Process
 #. Send out an email to `cyclus-dev` and `cyclus-users` to announce the release!
 
    
+.. This part has been commented, as it is required for the website, but the
+   person in charge of the release might not have the proper access to update the
+   Dory worker.  This should be automated when a merge is done on the master branch
+   a CI-hook should update the dory cloudlus server and relaunch the worker.
+   Moreover the cloudlus server is not directly related to Cyclus and depend on
+   the UW-Madison community but the website relies on it to host the
+   online cyclus calculation... (see
+   https://github.com/cyclus/cyclus.github.com/pull/227#pullrequestreview-21589660
+   discussion for more details.)
 
 .. #. Update ``Dory``'s ``Cyclus``/``Cycamore`` version and relaunch ``Dory`` worker.
    To do this you need a acces to the ``Dory`` server (if you don't please
--- a/source/cep/cep3.rst
+++ b/source/cep/cep3.rst
@@ -207,6 +207,19 @@ Release Process
       <https://developer.github.com/v3/repos/releases/#create-a-release>`_
 
 #. Update Conda-forge
+    - For each project, find the corresponding feedstock repository in the
+      conda-forge organization on github. For example, cyclus' feedstock is at
+      https://github.com/conda-forge/cyclus-feedstock
+
+    - In each project's feedstok, open up a PR which updates the
+      `recipe/meta.yaml` file with the new version number and the new SHA-256
+      value of the new version's tarball. See conda-forge documentation for more
+      or ask the feedstock maintainers for help. 
+
+    - Note that each feedstock must be accepted and the package uploaded to
+      anaconda.org (automatic) prior to accepting updates for the next feedstock
+      dependency. For example, cyclus must be fully updated before cycamore.
+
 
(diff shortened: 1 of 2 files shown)


Baaaaam:
@FlanFlanagan @gonuke @scopatz I don't know if/what needs to be done for conda when drafting a new release. Could one of you help me on that ?

source/cep/cep3.rst
```diff
@@ -243,6 +224,23 @@ Release Process
...
+   need a acces to the dory server, the go toolchain as well as cde installed on
+   your computer.
+    
+   .. code-block:: bash
+      $ git clone https://github.com/rwcarlsen/cloudlus.git
+      $ cd cloudlus
+      $ go install ./cmd/cloudlus
+      $ cd misc/fuelcycle.org
+      $ make
+      $ ssh dory.fuelcycle.org ‘mv cyc-cde.tar.gz cyc-cde.tar.gz_bkp’
```
scopatz: I think you want quotes here rather than forward ticks

scopatz:
Conda forge instructions are as follows

```rst
For each project, find the corresponding feedstock repository in the conda-forge organization on github. For example, cyclus' feedstock is at  https://github.com/conda-forge/cyclus-feedstock

In each project's feedstok, open up a PR which updates the `recipe/meta.yaml` file with the new version number and the new SHA-256 value of the new version's tarball. See conda-forge documentation for more or ask the feedstock maintainers for help. 

Note that each feedstock must be accepted and the package uploaded to anaconda.org (automatic) prior to accepting updates for the next feedstock dependency. For example, cyclus must be fully updated before cycamore.
```

scopatz:
please ping when this is ready

Baaaaam:
@gonuke I add the 'note' about the maintenance failing because of ABI

source/cep/cep3.rst
```diff
@@ -139,15 +139,35 @@ Release Process
       $ export CYCAMORE_DIR=/path/to/cycamore
       $ ./maintenence.sh -r -v X.X.X # X.X.X is *this* version
 
-#. Commit all changes for all projects
+    .. note:: 
+
+          If mainteance script fails because of an ABI failure that are cause by
```
gonuke: - maintenance
- "an ABI failure that is" or "ABI failures that are"
- caused

source/cep/cep3.rst
```diff
@@ -139,15 +139,35 @@ Release Process
...
+          a compiler update, you might want to accept them and procceed with the
```
gonuke: a compiler update (or other similar change caused by reasons other than code changes)

source/cep/cep3.rst
```diff
@@ -139,15 +139,35 @@ Release Process
...
+          release with those. to do so you need to generate the new symbols and
```
gonuke: "To do so,"

source/cep/cep3.rst
```diff
@@ -139,15 +139,35 @@ Release Process
...
+          commit them:
+          
+          #. First make sure those change can be ignore by email the dev-list
```
gonuke: - changes
- ignored
- emailing for discussion/approval

source/cep/cep3.rst
```diff
@@ -139,15 +139,35 @@ Release Process
...
+          
+          #. if the dev-list agreed to those changes, update the symbol and
```
gonuke: - agrees
- symbols

source/cep/cep3.rst
```diff
@@ -243,6 +257,24 @@ Release Process
 
 #. Commit all changes to ``cyclus.github.com`` and ``make gh-publish`` 
 
+#. Update Dory Cyclus/Cycamore version and relaunch Dory worker. To do this you
+   need a acces to the dory server, the go toolchain as well as cde installed on
+   your computer.
```
gonuke: These commands probably rely on certain state in your environment:
- you have the current Cyclus release in your path so cloudlus can find it
- other things???

source/cep/cep3.rst
```diff
@@ -243,6 +257,24 @@ Release Process
 
 #. Commit all changes to ``cyclus.github.com`` and ``make gh-publish`` 
 
+#. Update Dory Cyclus/Cycamore version and relaunch Dory worker. To do this you
+   need a acces to the dory server, the go toolchain as well as cde installed on
```
gonuke: - access
- let's put `dory`, `go`, and `cde` in backticks

Baaaaam:
@gonuke I addressed your comment.

@scopatz I think we are close to a final version, if you want to take a look :)
thx for the conda instructions btw!

source/cep/cep3.rst
```diff
@@ -120,12 +118,14 @@ Release Process
...
+   ``master``, ``develop``, and ``vX.X.X-release`` branches on ``upstream``.
 
-#. Bump the version in ``cyclus/src/version.h``, ``cycamore/src/version.h``, and
+#. Bump the version in ``cyclus/src/version.h.in``,
+   ``cycamore/src/cycamore_version.h.in``, and
    ``cymetric/setup.py``; commit the changes
 
-#. Perform maintainence tasks for all projects
+#. Perform maintenance tasks for all projects. The maintenance depends on `PyNe
+   <https://github.com/pyne/pyne.git>`_and Doxygen.
```
scopatz: pyne is spelled `PyNE`

source/cep/cep3.rst
```diff
@@ -193,46 +215,40 @@ Release Process
...
       $ git push upstream --delete vX.X.X-release
 
-#. Manually visit the github.com page for each project and mark the tags as releases
+#. Manually visit the github.com page for each project and mark the tags as releases.
 
     - This can be updated one day to use the Github `release API
       <https://developer.github.com/v3/repos/releases/#create-a-release>`_
 
+#. Update Conda-forge
+    - For each project, find the corresponding feedstock repository in the
```
scopatz: I think you need a single blank line above this one.

source/cep/cep3.rst
```diff
@@ -243,6 +259,29 @@ Release Process
 
 #. Commit all changes to ``cyclus.github.com`` and ``make gh-publish`` 
 
+#. Update 'Dory''s ``Cyclus``/``Cycamore`` version and relaunch ``Dory`` worker.
```
scopatz: I think the quotes on dory are weird. Also I am not sure that this should be a required maintenance task since only a few select people have SSH access to dory.  This would effectively limit the number of people who could possibly do a release. 

I my opinion, this documentation is good, but doesn't belong in CEP3

Baaaaam: Fair point, the problem here is the cloud server on fuelcycle.org, is hosted on dory. so it make sense to update dory on the same time we update the dory worker on the same time.

So I am not sure what is the best solution...

scopatz: I agree that it makes sense update dory at the same time. But I think that is the responsibility of the dory maintainers, not the release manager. 

One could imagine multiple instances of cloudlus put online by various entities. @FlanFlanagan and I plan on hosting cyclus server in some way publicly in the near future. I don't expect that anyone other than us should have to come in and update it. I think it makes more sense to have a system where those responsible for the compute are responsible for the maintenance. Just my opinion, though.

gonuke: tl;dr - if the website update is part of CEP3, then dory should be, too, but it's not clear who should be responsible

Dory is special because the website currently depends on it: users can submit input files to Dory for execution.  If that continues to be true, then perhaps we need to make sure there is always an identifiable person responsible for keeping it up to date and this instruction should be to notify that person to update it.  I agree that we don't necessarily want to have to share access to dory for all people who we want to manage a release.  That said - I think it requires a similar level of trust.

scopatz: My main concern is that this is effectively not automatable. I would like to write a release script that takes care of a lot of the minutia and burden of performing releases. I think this would be fine if this could be recast to opening up a PR with metadata updates that on merge by the maintainers will update dory.  Or something similar. If the burden is sufficiently low, it isn't a big  deal.

gonuke: We'll work on this idea of automation via CI hooks, etc

bam241:
@scopatz @gonuke I commented the `cloudlus` part (to keep a trace for when we want to do the automation via CI hooks)

source/cep/cep3.rst
```diff
@@ -193,58 +215,80 @@ Release Process
...
+      (``source/user/install_from_tarball.rst``)
 
 
 #. Commit all changes to ``cyclus.github.com`` and ``make gh-publish`` 
 
 #. Send out an email to `cyclus-dev` and `cyclus-users` to announce the release!
 
+   
+
+.. #. Update ``Dory``'s ``Cyclus``/``Cycamore`` version and relaunch ``Dory`` worker.
```
scopatz: Can you put another comment here please explaining why you commented it out. After that I think this is good to go. Thanks!

bam241:
@scopatz done !

scopatz:
Thanks @bam241

PR #232, opened 2017-02-13
few fix on CEP3

bam241:
- small syntax fix
- update modification date
- add BaM as an author (not sure what is the policy there, if the listed author are the orignal author only, or any one who updated it)


scopatz:
Thanks @bam241!

PR #233, opened 2017-03-31
Mbmore updates

mbmcgarry:
Added two new archetypes to the mbmore library.


scopatz:
Hi @mbmcgarry - thanks for putting this in. Looks like there are some conflicts, though

PR #238, opened 2017-05-17
update README

bam241:
Update sphinx requirement in readme

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ Dependencies
 
 Building the Cyclus website requires:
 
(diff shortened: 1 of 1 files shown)


scopatz:
I don't think this is the correct thing to do.  It is better to fix the bug than document allowable versions.

scopatz:
Thanks for putting this in though!

gonuke:
Is this still relevant?

scopatz:
I think not.  I think there was one version that was broken and then more recent versions of sphixed have fixed it.

bam241:
right closing!

PR #253, opened 2017-09-01
Simulation Orchestration Concern 

scopatz:
This is a modification to CEP2 that allows for something like Rickshaw

### DIFF
--- a/source/cep/cep2.rst
+++ b/source/cep/cep2.rst
@@ -211,7 +211,8 @@ The |cyclus| development team currently supports projects for these concerns:
   time, canonical supply chain entities)
 * `Cycamore`_ - Domain Models
 * `Cymetric`_ -  Analysis & Visualization
-* `Rickshaw`_ - Simulation orchestration.
+* `Cyclist`_ -  Analysis & Visualization (historical)
+* `Rickshaw`_ - Simulation orchestration
 
 The dependency graph for these projects is similar to the graph of the concerns.
 Figure 2 displays this graph along with other projects which implement these concerns.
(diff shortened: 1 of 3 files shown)


scopatz:
Note that this was discussed at the Cyclus community meeting in July.

source/cep/cep2.rst
```diff
@@ -137,58 +143,78 @@ which are pertinent to fuel cycle calculations:
...
 resource exchange put it with the domain models to be safe.
 
-The |cyclus| development team currently provides and supports three projects, 
-one for each concern:
+The |cyclus| development team currently supports projects for these concerns:
 
 * `Cyclus`_ - Dynamic Resource Exchange Simulation (individual actors, discrete
   time, canonical supply chain entities)
 * `Cycamore`_ - Domain Models
-* `Cyclist`_ -  Analysis & Visualization
```
gonuke: Should we somehow recognize `Cyclist` as an historical artifact?

scopatz:
> Related: do we have a CEP status for "Approved with pending modification", in part so that we can publish the site with the modified CEP? (although raises questions about publishing the diff, etc... hmmm)

Nope, there is nothing like that in CEP1 - but we could add it, if desired.

source/cep/cep2.rst
scopatz: Sure, I can add it back and put `(historical)` after it

scopatz: Done!

source/cep/cep2.rst
```diff
@@ -276,7 +308,9 @@ References and Footnotes
 
 .. _Cyclus: https://github.com/cyclus/cyclus
 .. _Cycamore: https://github.com/cyclus/cycamore
-.. _Cyclist: https://github.com/cyclus/cyclist2
+.. _Cymetric: https://github.com/cyclus/cymetric
+.. _Cyclist: https://github.com/cyclus/cyclist
```
gonuke: URL is `cyclist2`

scopatz: OOps Thanks! Fixed

bam241:
thx @scopatz !

everybody seems to agree... so I am merging

scopatz:
Thanks All!

PR #270, opened 2017-11-06
micro version number modification to cep3

scopatz:
A few lines that will make development easier going forward.


gonuke:
I support this change in practice, but recommend that we also add indications about this version naming policy to other places on the website for the benefit of users.

scopatz:
Thanks @gonuke - I have added some warnings around the site to this effect.

katyhuff:
:+1:

bam241:
thx @scopatz

PR #273, opened 2017-12-07
Fixing sidebar

FlanFlanagan:
This should fix the issue you were having @gonuke. It also fixes the dev call line. 

I does not fix the text issue on the sidebar however. Word wrapping needs to be added in some capacity but I'm having trouble forcing that in that section of the html.


bam241:
LGTM!

PR #282, opened 2018-03-27
Cep20 mod

FlanFlanagan:
First stab at this. Wasn't sure how to approach a modification of a CEP, but modifying the CEP was what suggested on the last dev call.

### DIFF
--- a/source/cep/cep20.rst
+++ b/source/cep/cep20.rst
@@ -17,7 +17,7 @@ progress through phases, including a building phase, a tick phase, a resource
 exchange phase, a tock phase, and a decommissioning phase. Phases are grouped in
 two categories: kernel phases and agent phases. Kernel phases have required or
 guaranteed actions that occur during them, and include the build, exchange, and
-decommission phases. Agent phases include the Tick and Tock phases, and allow
+decommission phases. Agent phases include the Tick, Tock, and Decision phases, and allow
 agents to inspect simulation state and update their own state.
 
 Motivation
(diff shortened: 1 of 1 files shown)


source/cep/cep20.rst
```diff
@@ -178,8 +178,31 @@ with the exception of the core's handling of agent entry/exit
 registration. *Cycamore* modules that deal with agent entry/exit will have to be
 redesigned to incorporate the new execution stack.
 
+Amendment 1. Decision Phase (Post Decommission)
+===================================
```
scopatz: FORMATTING: the line underneath needs to be at least as long as the title.

source/cep/cep20.rst
```diff
@@ -178,8 +178,31 @@ with the exception of the core's handling of agent entry/exit
 registration. *Cycamore* modules that deal with agent entry/exit will have to be
 redesigned to incorporate the new execution stack.
 
+Amendment 1. Decision Phase (Post Decommission)
```
scopatz: I don't think that this should be its own amendment section, but rather integrated into the text of the document.

scopatz:
Thanks for putting this in @FlanFlanagan! The text looks good to me, but I would like to see it a bit more integrated into the document

scopatz:
Also, I like the name "decision phase"

FlanFlanagan:
Okay! I'll integrate it more. Wasn't sure how to go about approaching it so thanks for the suggestion!

FlanFlanagan:
Modified again. I also moved the Decision phase to directly after the Tock phase as I originally misread the requirements. It's that user phases have to bracket kernal phases, not the other way around. I think this is better anyway as it allows for a decommission condition to occur on the same time step as the decision.

FlanFlanagan:
Ordering of the tock calls is not manageable by the user or agents. Therefore if the institution is agent2 it's decision making will occur before agents spawned by that institution. So any behavior they make during tock will be unknown. 

I shifted it because, I reread the convention and it's just that kernal phases must be surrounded by agent phases. Not the other way around. This positioning also allows users to plan a decommission on the current timestep, as opposed to leaving an agent up for the next time step before the decommission order goes though.

gonuke:
Perhaps this ordering notion could/should be added to the CEP.  It's subtle but important, and easily forgotten, as evidenced by my question....

Given this issue of ordering in any given phase, there is an obvious question about whether you will always need one more phase to deal with the fact that the actions (or decisions) of the previous phase cannot be ordered.  It may be useful to draw a line in a sand here?

FlanFlanagan:
I can add some information that will explain the decision to put it where it is. 

As to your other point, I think you're right. We can continue to add phases on top of phases, we need to draw the line somewhere. I personally think this is the right place to draw it, but I'm open to other ideas.

bam241:
thx @FlanFlanagan

PR #296, opened 2019-10-15
[WIP] Reboot of CEP7 development

gonuke:
In light of possible expansions and changes to the Cyclus community, I am rebooting the consideration of CEP7.

### DIFF
--- a/source/cep/cep7.rst
+++ b/source/cep/cep7.rst
@@ -97,6 +97,13 @@ Usability
 Code of Conduct
 ================
 
+Some other related codes of conduct during the development of our own:
+
+* http://contributor-covenant.org/
+* http://www.ubuntu.com/about/about-ubuntu/conduct
+* http://www.apache.org/foundation/policies/conduct.html
+* https://www.mozilla.org/en-US/about/governance/policies/participation/
+
 
 Change Control Process
 ========================
--- a/source/cep/cep7.rst
+++ b/source/cep/cep7.rst
@@ -1,0 +1,242 @@
+CEP 7 - Cyclus Community Constitution
+*************************************
+
+:CEP: 7
+:Title: Cyclus Community Constitution
+:Last-Modified: 2015-08-18
+:Author: Paul Wilson
+:Status: Active
+:Type: Process
+:Created: 2015-08-18
+
+Abstract
+========
+
+This CEP is intended to define fundamental principles of the |Cyclus|
+organization and a basis for governance within this community.  Prior to this
+document, governance was ad-hoc within a single institution.  As the
+individual and institutional relationships expand, such a document is
+considered important to maintain the health of the community.
+
+Mission
+=======
+
+The mission of the |Cyclus| organization is to foster the ongoing development
+of the open source |Cyclus| ecosystem as both a research platform and a user
+tool for the study of dynamic nuclear fuel cycles.
+
+Defintions
+==========
+
+|Cyclus| Ecosystem
+-------------------
+
+The |Cyclus| ecosystem is an organic collection of individuals, institutions,
+and their contributions to nuclear fuel cycle modeling and analysis within the
+|Cyclus| framework.  Those contributions may be distinct projects, individual
+lines of code, or some combination, and they may be owned individually,
+collectively, institutionally, or in some combination.
+
+|Cyclus| Organization
+----------------------
+
+The |Cyclus| organization (hereafter, "the Organization") is the subset of the
+|Cyclus| ecosystem that is managed and owned collectively under the |Cyclus|
+Github organization.  This document is designed to provide a basis for
+governance within this organization.  Membership in this organization is
+governed by a process defined below.
+
+Council of Principal Investigators
+-----------------------------------
+
+????
+
+Design Philosophy
+==================
+
+The fundamental design philosophy of |Cyclus| provides a basis for guiding the
+development of the components owned by the Organization and resolving
+technical disputes that may arise during such development.  Such a philosophy
+is embodied in the following set of principles.  Like any such system, the
+relative value placed in any single principle may not be universally held by
+all members of the Organization.
+
+Flexibility
+------------
+
+The foundational principle of |Cyclus| is to provide flexibility in both the
+range of nuclear fuel cycles that can be modeled and the range of analyses and
+use cases that may be applied to those cycles.  
+
+Open
+-----
+
+The Organization is commited to an open source development paradigm
+
+Scalability
+------------
+
+
+Stability
+----------
+
+Stability of |Cyclus| components is an important prerequisite for attracting a
+community of developers and users.  A stable |Cyclus| kernel is required to
+attract other archetype developers and stability in the other components is
+required to attract end users.
+
+Sustainability
+---------------
+
+
+Usability
+----------
+
+
+
+Code of Conduct
+================
+
+
+Change Control Process
+========================
+
+The primary context for interaction among members of the Organization is the
+discussion and review of changes being made to the software.  Such changes are
+enabled by the pull request features of Github technology that hosts the
+primary software repositories.
+
+1. A developer or team of developers makes changes in a Github branch.
+2. Once a cohesive set of changes has been completed, a pull request is
+   generated that offers those changes as an ammendment to the primary branch.
+3. All members of the Organization are welcome to comment on all technical
+   aspects of the offered changes, including specfically:
+   * style
+   * technical correctness
+   * adherence to the design philosophy
+4. At least one member of the Organization that was not involved in the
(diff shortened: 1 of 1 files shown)


jbae11:
@gonuke Thank you for putting this together! If I understand correctly, there is now going to be a [process to become a Cyclus developer](https://github.com/cyclus/cyclus.github.com/pull/296/files#diff-728a998fac47df94ed5afc9e6b76e8b1R47)?

gonuke:
@jbae11 - This is not about a process to become a developer - we don't need an more obstacles there.

This is about a process to govern a community of those who effectively volunteer their effort to that community.

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+Escalation to this response implies that previous attempts to resolve the
+situation were unsuccessful and/or that there is an emerging pattern of
+transgression.  The PI are included both to make them aware of the situation
+and to invite them to take independent action.  This response also implies an
+explicit judgment and hence imposes stigma on the recipient.
+
+Any member of the community is empowered to choose this response at their own
+discretion.
+
+4. Greivance
```
jbae11: Grievance

jbae11 (approved):
Thanks!

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+|Cyclus| Organization
+----------------------
+
+The |Cyclus| organization (hereafter, "the Organization") is the subset of the
+|Cyclus| ecosystem that is managed and owned collectively under the |Cyclus|
+Github organization.  This document is designed to provide a basis for
+governance within this organization.  Membership in this organization is
+governed by a process defined below.
+
+Council of Principal Investigators
```
munkm, line 49: Some thoughts about the council: 
* If a community manager exists, their perspective would be really useful in the council. Maybe it would be helpful to reframe this council as a "leadership" council that is encoded to have a set of PIs on it
* The perspective of a person who is exclusively a user, and not a developer, can be helpful at a top level council making decisions. Would it be helpful to add in a role for this council that fills this representation?

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+   * style
+   * technical correctness
+   * adherence to the design philosophy
+4. At least one member of the Organization that was not involved in the
+   development of the changes must approve the changes.
+5. Any member of the Organization can block the adoption of changes on
+   technical grounds.
+6. Changes are adopted when at least one member of the Organization has
+   approved the changes and no members of the Organization are blocking the
+   changes.
```
munkm, line 130: For more substantive changes (new features maybe?) it might be worth adding in a requirement that a PR is open for a period of time (perhaps two weeks?) before merging is allowed. That way community members have time to weigh in even if they aren't able to attend a development meeting where new features are getting pushed forwards.

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
+CEP 7 - Cyclus Community Constitution
+*************************************
+
+:CEP: 7
+:Title: Cyclus Community Constitution
+:Last-Modified: 2015-08-18
```
munkm, line 6: If this CEP gets updated as a result of this meeting (or the hackathon) then we should update this (or do we have a bot that handles it?)

```suggestion
:Last-Modified: 2024-03-21
```

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+:Author: Paul Wilson
```
munkm, line 7: Do reviewers count as authors for CEPs?

katyhuff:
It appears I've been ignoring this review request for 5 years. Madicken (@munkm ) seems to have done a recent review. Is my review still needed?

gonuke:
Since CEP7 is about governance, I think your input would still be valuable

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+
+(should these be in their own CEP)
+
+1. Substantive changes to the work of others should require those others to
+   comment during the review process.
+2. Individual pull requests should be as small as possible to facilitate
+   timely review.
+3. Before embarking on long time-intensive changes, it is wise to collect the
+   opinion of the |Cyclus| community on the value of such changes, generally
+   through the mailing list.
```
munkm, line 143: I like the idea of having a development best practices CEP (I think this is what is commented on in l135?). This section isn't specific to governance and broad decision making, but it is something we want to encourage (or require?) community members to practice.

source/cep/cep7.rst
katyhuff: Two thoughts on these two good thoughts. 

1) With respect to the community manager, I think this makes sense in principle. But, in practicality: No community manager exists at this time, and it's hard for me to imagine a Cyclus community manager that isn't PI (or Co-PI) of any funded work. So, if a Cyclus community manager exists, that person is probably also already included as a PI. My tendency would be to leave this as-is until there is, someday, a non-PI community manager. 
2) With respect to the person who is exclusively a user -- again, great idea in principle, but in practice, I see some issues. A single user is never representative of all users. And, a user who is exclusively a user but also sufficiently interested to be part of a council of PIs, is very likely, at least in the Cyclus universe, to become a developer fairly quickly.  I would suggest that the council should simply be expected to to gather population-level input from the whole exclusive-user population on a routine basis, especially when that input is useful for decisions.

source/cep/cep7.rst
katyhuff: I'm a little concerned about this 2 week timeout. Sometimes event big changes need to happen quickly. I'm sure that most PRs stay open at least that long anyway, but when one needs to be merged quickly, it would be a shame for this kind of rule to stand in its way. There's nothing fatal about merging something that someone may wish fixed later -- it's not ideal, but it doesn't become set in stone. If someone's PR creates a sufficiently serious issue for someone that they would have otherwise stood in the way of it being pulled -- they can always create an issue and PR to fix it (or even revert it). There's no deadline one that.

katyhuff:
I think this looks good. 
As long as the modified date is up to date, I approve of this.

katyhuff (approved):
As long as the last-modified date is correct, I approve.

source/cep/cep7.rst
gonuke: I don't think there's a bot... We probably need to do that manually and the clock is still ticking.  Maybe I should look into a bot...

source/cep/cep7.rst
gonuke: It seems like we still need to define how membership on the Council of PIs is formally defined (given the `????` below :) ).  I know that we once discussed it being PIs who are explicitly funded to work on Cyclus, but I wonder if that's complete enough?  We could always start with that and go from there.

gonuke:
Alas... the document is actually incomplete, as written, and I'd welcome PR's or other modes to suggest improvements, particularly to those incomplete places.  For example:

* we haven't defined who makes up the Council of PIs
* there is no elaboration on the design philosophies points on Scalability, Sustainability, Usability
* do we need to define a formal grievance process?

source/cep/cep7.rst
abachma2: We do have a CEP on PRs into Cyclus (https://fuelcycle.org/cep/cep6.html). We'll need to make sure that this CEP does not conflict with what we have here -- or have this CEP point to that one regarding PRs.

source/cep/cep7.rst
abachma2: We do have a CEP specifically for archetype development best practices: https://fuelcycle.org/cep/cep5.html. If we made a best development practices CEP, would we want this CEP to point to that one? Let's make sure nothing is contradicting between CEPs.

source/cep/cep7.rst
```diff
@@ -0,0 +1,267 @@
...
+3. Monitored Flags
+-------------------
+
+Clearly an escalation from the previous responses, this follows the same
+pattern as other flags, but explicitly includes one or more members of the
+Council of PIs, as appropriate.
+
+Escalation to this response implies that previous attempts to resolve the
+situation were unsuccessful and/or that there is an emerging pattern of
+transgression.  The PI are included both to make them aware of the situation
```
abachma2, line 247: There is a grammar mismatch here: PI is singular, are is plural. Are you intending "PI" as in the council of PIs? Or just one of the PIs on the council?

source/cep/cep7.rst
abachma2: Honestly, I think I'm probably the closest we have to a community manager right now, although I'm not sure if we have a formal definition for that anywhere. Also, how would this council of PIs interact with the BDFP discussed in [CEP 1](https://fuelcycle.org/cep/cep1.html)? Do we have an outline of the council's responsibilities?

PR #297, opened 2019-10-30
CEP27: toolkit Snippet

bam241:
This is a suggestion to improve, ease and homogenize an the usage of toolkit capability in archetypes.

This idea is to limit as much as possible the need for implementation in a newly develop archetypes to avoid divergence in the implementation and the convention by providing a snippet that one can simply includes in the archetypes class definition header (inside the core definition of the class) to ensure variable name are identical.

Example can be found in:
- cyclus/cyclus#1510 for the snippet implementation
- cyclus/cycamore#502 for the corresponding archetypes implementations

### DIFF
--- a/source/cep/cep27.rst
+++ b/source/cep/cep27.rst
@@ -27,16 +27,18 @@ in the implementation and the usage, across multiple archetypes.
 Toolkit Implementation
 ======================
 
-Each |Cyclus| toolkit component will contain 3 different files: 2 for the definition of the
-snippet C++ class (``cpp`` and header) that allows the use of the capabilities, and
-optionally to register its values in the output database, a snippet simplifying
-the integration of the feature in a newly develop archetypes.
+Each |Cyclus| toolkit component will contain 3 different files:
+- 2 for the definition of the snippet C++ class (``cpp`` and header) that allows
+  the use of the capabilities, and optionally to register its values in the
+  output database,
+- a snippet definition file used to simplify the implementation and ensure
+  consistency accross its integration in the different archetypes.
 
-The snippet file with then be included in the header part of the archetypes
+The snippet definition file will then be included in the header part of the archetypes
 class declaration as: ``#include toolkit/my_snippet.cycpp.h``
-
 (The use of the ``cycpp.h`` has been chosen to allow syntax highlighting and
 inform developers that this is not a standard C++ header)
+
 The snippet file, will contain the declaration of all the variables required
 to use the capabilities class:
 
@@ -72,8 +74,8 @@ have to be done:
 2. Add the proper default initialization of the variable required for the
    snippet.
 
-3. In the ``Archetype::EnterNotify()``, assign the Snippet with value
-   corresponding to the user input.
+3. In the ``Archetype::EnterNotify()``, initialise the toolkit class member
+   variables with variables.
 
 4. (optional) If required, call the ``RecordSnippet()`` method when necessary during the
    Archetype operation.
@@ -91,4 +93,57 @@ develop to skip the assignation of the value in the inherited class in the
 ``EnterNotify``...
 
 Otherwise behavior would be very similar.
+
+Exemple:
+========
+
+``toolkit/my_snippet.cycpp.h``:
+.. highlight:: c
+    cyclus::toolkit::Position coordinates;
+    
+    #pragma cyclus var { \
+        "default": 0.0, \
+        "uilabel": "Geographical latitude in degrees as a double", \
+        "doc": "Latitude of the agent's geographical position. The value should " \
+           "be expressed in degrees as a double." }
+    double latitude;
+    // required for compilation but not added by the cycpp preprocessor...
+    std::vector<int> cycpp_shape_latitude;
+
+    #pragma cyclus var { \
+           "default": 0.0, \
+           "uilabel": "Geographical longitude in degrees as a double", \
+           "doc": "Longitude of the agent's geographical position. The value should" \
+           "be expressed in degrees as a double." }
+    double longitude;
+    // required for compilation but not added by the cycpp preprocessor...
+    std::vector<int> cycpp_shape_longitude;
+
+
+``my_archetype_example.h``:
+.. highlight:: c
+    class fun_archetype : public cyclus::facility{
+        public:
+        [...]
+        private:
+        [...]
+        #include "toolkit/my_snippet.cycpp.h"
+
+``my_archetype_example.cpp``:
+.. highlight:: c
+    fun_archetype::fun_archetype(cyclus::Context* ctx): 
+        cyclus::facility(ctx),
+        var1(0.0),
+        var2(0.0),
+        ...,
+        coordinates(0,0), //coordinates constructor (toolkit feature class)
+        longitude(0), //snippet variables added with "my_snippet.cycpp.h"
+        latitude(0) //snippet variables added with "my_snippet.cycpp.h"
+    {}
+    [..]
+    void Storage::EnterNotify() {
+        coordinates.set_position(latitude, longitude);
+        coordinates.RecordPosition(this);
+        [...]
+        }
 
--- a/source/cep/cep27.rst
+++ b/source/cep/cep27.rst
@@ -26,7 +26,7 @@ feature inside multiple archetypes.
 .. _agent-spec-docs:
 
 Agent Spec
-===========
+==========
 
 Agents will be uniquely identified by 3 components together referred to as an
 "agent spec":
--- a/source/cep/cep27.rst
+++ b/source/cep/cep27.rst
@@ -1,5 +1,5 @@
-CEP 27 - Agent Toolkit Capabilities
-***********************************
+CEP 27 - Toolkit Capabilities Injection into an Archetype
+*********************************************************
 
 :CEP: 27
 :Title: Agent Toolkit Capabilities
--- a/source/cep/cep27.rst
+++ b/source/cep/cep27.rst
@@ -5,7 +5,7 @@ CEP 27 - Agent Toolkit Capabilities
 :Title: Agent Toolkit Capabilities
 :Last-Modified: 2019-10-28
 :Author: Baptiste Mouginot <mouginot.baptiste@gmail.com>
-:Status: Pending
+:Status: Active
 :Type: Standards Track
 :Created: Baptiste Mouginot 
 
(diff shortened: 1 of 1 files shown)


source/cep/cep27.rst
```diff
@@ -0,0 +1,94 @@
...
+  cycpp_shape_myvariable0`` to be declared for each of the decorated variable.
+  ``cycpp preprocessor`` is not able at the time to add them automatically for
+  included files.
+
+
+The main interest of this include method would be to insure consistency across
+archetypes using the same toolkit capability requiring user input, avoiding 1
+set of input syntax per archetypes for the same capability.
+
+If the toolkit features needs to capabilities to write in the output database a
```
gonuke:
```suggestion
If the toolkit features needs the capabilities to write in the output database a
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,94 @@
...
+Archetypes Integration
+======================
+
+When the capability is integrated in an Archetype the following implementations
+have to be done:
+
+1. Include the snippet in the class header core: ``#include
+   toolkit/my_snippet.cycpp,h``.
+
+2. Add the proper default initialisation of the variable required for the
```
gonuke:
```suggestion
2. Add the proper default initialization of the variable required for the
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,94 @@
...
+   snippet.
+
+3. In the ``Archetype::EnterNotify()``, assign the Snippet with value
```
gonuke: Do you really assign the "Snippet" with a value???

bam241:
Cc @yarden-livnat

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
+CEP 27 - Agent Toolkit Capabilities
```
gonuke: Probably want a slightly more precise/descriptive title since this CEP is about how to inject/integrate the toolkit capability into an archetype.

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
...
+:Status: Active
+:Type: Standards Track
+:Created: Baptiste Mouginot
+
+
+Abstract
+========
+
+The |Cyclus| toolkit is designed to easily implement specific capabilities in
+newly developed archetypes, such as trading policy, commodity producers... To
```
gonuke:
```suggestion
newly developed archetypes, such as trading policy, commodity producers. To
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
...
+add characteristics to archetypes such as Position or Metadata, the actual
```
gonuke: Not sure it makes sense to refer to specific toolkit capabilities.  If we do, then:
```suggestion
add characteristics to archetypes such as `Position` or `Metadata`, the actual
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
...
+implementation method is very verbose, where in each archetypes one needs to add
+the new specification in the arcehtype header, then assign it and use it in the
+cpp file, with no guarantee on the consistency in the variable naming and
```
gonuke: This sentence kind of runs on... Maybe consider making it a couple of sentences.

gonuke: Also - perhaps this is more representative (few things are every guaranteed :smile: ):
```suggestion
cpp file, and the developer must ensure the consistency in the variable naming and
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
...
+
+The snippet file, will contain the declaration of all the variables required
+to use the capabilities class:
+
+- the definition of the capability class as a member variable.
+
+- (optional) if the capability requires/allows variable input from users,
+  standard |Cyclus| member variable declaration with variable ``#pragma`` is
+  required. In addition, to the standard variable declaration and the
+  ``#pragma`` the method also require a ``std::vector<int> cycpp_shape_myvariable``
```
gonuke: Are these declarations in the snippet file?

source/cep/cep27.rst
```diff
@@ -0,0 +1,148 @@
...
+archetype class declaration in addition to simply including the snippet at the
+right place.
+This may also lead to confusion, as one can believe that the user input value
+for variable are passed in the constructor of the class and might lead the
+develop to skip the assignation of the value in the inherited class in the
+``EnterNotify``...
+
+Otherwise behavior would be very similar.
+
+Exemple:
```
gonuke:
```suggestion
Example:
```

gonuke: Can we include an example for inheritance as well as class member versions?

source/cep/cep27.rst
bam241: yes it should be included in the `toolkit/my_snippet.cycpp.h` file

source/cep/cep27.rst
```diff
@@ -0,0 +1,202 @@
...
+:Status: Active
+:Type: Standards Track
+:Created: Baptiste Mouginot
+
+
+Abstract
+========
+
+The |Cyclus| toolkit is designed to easily implement specific capabilities in
+newly developed archetypes, such as trading policy, commodity producers. To
```
gonuke:
```suggestion
newly developed archetypes, such as a trading policy, commodity producers, etc. To
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,202 @@
...
+add characteristics to archetypes such as `Position` or `Metadata`, the actual
+implementation method is very verbose, where in each archetypes one needs to add
+the new specification in the arcehtype header, then assign it and use it in the
+cpp file, and the developer must ensure the consistency in the variable naming and
+implementation across multiple archetypes.
+This CEP explains how to use snippets to simplify and maintain consistency
```
gonuke:
```suggestion
This CEP explains introduces the concept of snippets to simplify and maintain consistency
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+``my_archetype_example.cpp``:
+.. highlight:: c
+    fun_archetype::fun_archetype(cyclus::Context* ctx):
+        cyclus::facility(ctx),
+        var1(0.0),
+        var2(0.0),
+        ...,
+        coordinates(0,0), //coordinates constructor (toolkit feature class)
```
gonuke: This initializer is not necessary/valid because the snippet in this example doesn't include the `coordinates` variable.

source/cep/cep27.rst
```diff
@@ -0,0 +1,202 @@
...
+
+Example:
+========
+
+
+Without Inheritance:
+--------------------
+``toolkit/my_snippet.cycpp.h``:
+.. highlight:: c
+    cyclus::toolkit::Position coordinates;
```
gonuke: Is this only in the version without inheritance?  Is this the only difference between the **with** and **without**?

source/cep/cep27.rst
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+``my_archetype_example.cpp``:
+.. highlight:: c
+    fun_archetype::fun_archetype(cyclus::Context* ctx):
+        cyclus::facility(ctx),
+        var1(0.0),
+        var2(0.0),
+        ...,
+        coordinates(0,0), //coordinates constructor (toolkit feature class)
```
gonuke: if we use C++11 can we initialize these in the snippet?

source/cep/cep27.rst
bam241: it also change in the implementation of the archetype

source/cep/cep27.rst
bam241: yes that's true, this will be way better !

bam241: I'll try this in my Position PR and see how it does work out!

bam241: it seems to be working, I updated both the PR and the CEP accordingly

source/cep/cep27.rst
```diff
@@ -0,0 +1,180 @@
...
+           "default": 0.0, \
+           "uilabel": "Geographical longitude in degrees as a double", \
+           "doc": "Longitude of the agent's geographical position. The value should" \
+           "be expressed in degrees as a double." }
+    double longitude = 0;
+    // required for compilation but not added by the cycpp preprocessor...
+    std::vector<int> cycpp_shape_longitude;
+
+``my_archetype_example.h``:
+.. highlight:: c
```
gonuke: Should we add the `#include` line here for the toolkit class. e.g.
```
#include 'toolkit/Position.h'
```
This will help make it clear that both files need to be included.

source/cep/cep27.rst
```diff
@@ -0,0 +1,180 @@
...
+:Type: Standards Track
+:Created: Baptiste Mouginot
+
+
+Abstract
+========
+
+The |Cyclus| toolkit is designed to easily implement specific capabilities in
+newly developed archetypes, such as a trading policy, commodity producers, etc. To
+add characteristics to archetypes such as `Position` or `Metadata`, the actual
```
gonuke: This is still a long, awkward sentence.

source/cep/cep27.rst
gonuke: I mean the only difference in the snippet.  It would be great if the same snippet could be used in both modes.

source/cep/cep27.rst
bam241: Yes I forgot about the `#include 'toolkit/Position.h`,

I am wondering if it would work if we put it inside the `toolkit/my_snippet.cycpp.h` ?

gonuke: definitely not for inheritance

bam241: Good point !

source/cep/cep27.rst
```diff
@@ -0,0 +1,184 @@
...
+
+
+Abstract
+========
+
+The |Cyclus| toolkit is designed to easily implement specific capabilities in
+newly developed archetypes, such as a trading policy, commodity producers, etc.
+To add characteristics to archetypes such as `Position` or `Metadata`, the
+actual implementation method is very verbose. It relies on adding the new
+specification in the arcehtype header, assignng it and use it in the cpp
```
gonuke:
```suggestion
specification in the arcehtype header, assigning it and use it in the cpp
```

source/cep/cep27.rst
```diff
@@ -0,0 +1,187 @@
...
+
+
+Archetypes Integration
+======================
+
+When the capability is integrated in an Archetype the following implementations
+have to be done:
+
+1. Include toolkit class header in in the class header:
+   ``#include 'toolkit/my_feature.h'``.
```
bam241: @gonuke Just added this, as you pointed out it was missing.

gonuke:
I still wonder if there is a way for the same snippet to be used for inheritance and membership?

bam241:
> I still wonder if there is a way for the same snippet to be used for inheritance and membership?

that would be nice ! 

maybe an other way to do that would be to let `cycpp` do that using `#pragmas` ?

gonuke:
Maybe we don't include the member variable in the snippet?  The user needs to define the name for that member variable?  That relaxes one kind of imposed consistency, but maybe that variable name is purely internal and doesn't need to be consistent... just a thought

bam241:
> Maybe we don't include the member variable in the snippet? The user needs to define the name for that member variable? That relaxes one kind of imposed consistency, but maybe that variable name is purely internal and doesn't need to be consistent... just a thought

I liked the snipped for the variable exposed in the `cyclus` input to ensure those one are the same

source/cep/cep27.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Example:
+========
+
+
+Without Inheritance:
+--------------------
+``toolkit/my_feature_snippet.cycpp.h``:
+.. highlight:: c
+    cyclus::toolkit::Position coordinates(0,0);
```
gonuke: If you remove only this line (not exposed in input), then the snippet can be used in both modes, and developer chooses name of member variable.

In fact, that might be better, because otherwise developer has to know name of member variable defined in the snippet.

bam241: this is correct, (and I like the versatility it provides)
I guess it is a balance between versatility and amount of work on the developer side:
if we only support without-inheritance everything should be includable in the snippet.
if we want to support both, developers have to add:
- `#include 'toolkit/my_toolkit.h`
- inheritance or feature element
- `#include `toolkit/snippet.cycpp.h` in the class core.

(this is also why without inheritance has my (slight) preference.)

bam241: the good thing with removing the variable it would allow both with and without inheritance

gonuke: Even without inheritance, I'm not sure we want to assume that the member variable is always defined by the snippet... what if you want two positions?

bam241: sure but then what is the use of the snippet ? because some member variable are defined in the snippet (at least the one expose in the input.)

I don't see how to use the snippet with 2 positions.

gonuke: This might not be the best example of including more than one instance.  In addition, there may be a need to initialize each instance which gets complicated again.  Nevertheless, I think asking the developer to explicitly declare their member variable is not too much to ask.

We should open this up to additional discussion, e.g. on the mailing list.

katyhuff:
Just a couple little grammar comments suggested above -- otherwise looks good to me.

source/cep/cep27.rst
gonuke: I still think there is a benefit to having a single snippet file and then letting user choose the variable name in the non-inheritance case.  We can include that in the example in the documentation.

Once we get that resolved, and clean up this CEP, I think I'm ready to approve/merge this and the follow-up PR's in the code repos.

gonuke (approved):
I think this is ready to move forward

PR #303, opened 2020-03-29
force updating/creating local/master from upstream/master on publish

bam241:
Added this, to ensure `make xxx-publish` is using an up-to-date version of `master` and `source` to build and publish the last version of the website.

Context:
Earlier today, I try to publish the new version of the website, the process requires you to have a local `master` branch, which I did not have... So I created it from `source` not from `upstream/master`.
I then ran the build and publish process (which updates and push a new version of `master`) but as my `master` was a copy of `source` and not `master` it did not include the pre-built files required for the website... and so it broke the website...

This should ensure that nobody make this kind of mistake in the future:
the `git check -B master upstream/master` creates a new `master` branch locally if it did not already exist, or resets it to `upstream/master` if existing...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,8 @@ gh-publish-only:
 	rsync -a $(BUILDDIR)/.* .
 	git add -f `(cd $(BUILDDIR); find . -type f; cd ..)`
 	rm -rf $(GH_SOURCE_DIR) $(BUILDDIR)
-	#git commit -m "Generated $(GH_PUBLISH_BRANCH) for `git log $(GH_SOURCE_BRANCH) -1 --pretty=short --abbrev-commit`" && git push --force $(GH_UPSTREAM_REPO) $(GH_PUBLISH_BRANCH)
-	#git checkout $(GH_SOURCE_BRANCH)
+	git commit -m "Generated $(GH_PUBLISH_BRANCH) for `git log $(GH_SOURCE_BRANCH) -1 --pretty=short --abbrev-commit`" && git push --force $(GH_UPSTREAM_REPO) $(GH_PUBLISH_BRANCH)
+	git checkout $(GH_SOURCE_BRANCH)
 
 gh-publish:
 	make clean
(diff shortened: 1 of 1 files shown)


bam241:
I also change which `source` branch is pulled into the master branch. now it pulls the `upstream/source` and not the local one....

Makefile
```diff
@@ -78,12 +78,13 @@ gh-preview html:
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)."
 
 gh-publish-only:
-	git checkout $(GH_PUBLISH_BRANCH)
-	git checkout $(GH_SOURCE_BRANCH) -- $(GH_SOURCE_DIR)
+	git fetch $(GH_UPSTREAM_REPO)
+	git checkout -B $(GH_PUBLISH_BRANCH) $(GH_UPSTREAM_REPO)/$(GH_PUBLISH_BRANCH)
+	git checkout $(GH_UPSTREAM_REPO)/$(GH_SOURCE_BRANCH) -- $(GH_SOURCE_DIR)
```
gonuke: Don't need this line anymore

Makefile
```diff
@@ -78,12 +78,13 @@ gh-preview html:
...
 	git reset HEAD
 	rsync -a $(BUILDDIR)/* .
 	rsync -a $(BUILDDIR)/.* .
-	git add `(cd $(BUILDDIR); find . -type f; cd ..)`
+	git add -f `(cd $(BUILDDIR); find . -type f; cd ..)`
 	rm -rf $(GH_SOURCE_DIR) $(BUILDDIR)
```
gonuke:
```suggestion
	rm -rf  $(BUILDDIR)
```

Makefile
bam241: are you sure, this is suppose to pull the `source` branch into the `master` `source` folder

gonuke: Yep.  I wrote these files originally.  We used to run `make clean` and `make html` immediately following this line, to ensure we got a fresh build of the site.  In the subsequent lines, the only files that we add to this branch are the ones in the root directory that have the same names as the files in `$(BUILDDIR)` which is a sibling of the source dir.   (see the `Makefile` in the CNERG website repo for reference)

When we modularized and made the `gh-pubish-only` we moved the `make clean` and `make html` to occur before all of this repo manipulation, so it doesn't matter what source branch we pull here.

bam241:
I made the fix. let me know if you want more change there.

gonuke:
Thanks @bam241

PR #304, opened 2020-03-30
importing new Makefile from source

bam241:
this propagates the changes in Makefile from `source` branch


gonuke:
Thanks for the follow-up @bam241

PR #332, opened 2024-02-16
Fix commit in gh-publish-only

abachma2:
When trying to publish changes made in from the last few PRs merged, I got an error message that 
```
git commit -m "Generated master for `git log source -1 --pretty=short --abbrev-commit`" && git push --force upstream master
fatal: ambiguous argument 'source': both revision and filename
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This change should fix this error, removing the ambiguity in the source branch and the source directory and allow us to publish the changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ gh-publish-only:
 	rsync -a $(BUILDDIR)/.* .
 	git add -f `(cd $(BUILDDIR); find . -type f; cd ..)`
 	rm -rf  $(BUILDDIR)
-	git commit -m "Generated $(GH_PUBLISH_BRANCH) for `git log $(GH_SOURCE_BRANCH) -1 --pretty=short --abbrev-commit`" && git push --force $(GH_UPSTREAM_REPO) $(GH_PUBLISH_BRANCH)
+	git commit -m "Generated $(GH_PUBLISH_BRANCH) for `git log $(GH_SOURCE_BRANCH) -- -1 --pretty=short --abbrev-commit`" && git push --force $(GH_UPSTREAM_REPO) $(GH_PUBLISH_BRANCH)
 	git checkout $(GH_SOURCE_BRANCH)
 
 gh-publish:
--- a/source/cyclusagent.py
+++ b/source/cyclusagent.py
@@ -25,8 +25,6 @@ except ImportError:
 from docutils import nodes
 from docutils.parsers.rst import Directive
 from docutils.parsers.rst import directives
-#from docutils.parsers.rst.roles import set_classes
-#from docutils.transforms import misc
 from docutils.statemachine import ViewList
 
 from sphinx.util.nodes import nested_parse_with_titles


abachma2:
The steps in the Makefile are:

```
make clean
make html
git fetch $(GH_UPSTREAM_REPO)
git checkout -B $(GH_PUBLISH_BRANCH) $(GH_UPSTREAM_REPO)/$(GH_PUBLISH_BRANCH)
git reset HEAD
rsync -a $(BUILDDIR)/* .
rsync -a $(BUILDDIR)/.* .
git add -f `(cd $(BUILDDIR); find . -type f; cd ..)`
rm -rf  $(BUILDDIR)
git commit -m "Generated $(GH_PUBLISH_BRANCH) for `git log $(GH_SOURCE_BRANCH) -1 --pretty=short --abbrev-commit`" && git push --force $(GH_UPSTREAM_REPO) $(GH_PUBLISH_BRANCH)
	git checkout $(GH_SOURCE_BRANCH)
```

`$(BUILDDIR)` is `./gh-build`, `${GH_PUBLISH_BRANCH)` is `master`, `$(GH_SOURCE_BRANCH)` is `source`, and `$(GH_UPSTREAM_REPO)` is `upstream`

So it does do all of the steps you described. The master branch does have a directory names source. I don't remember having issues with this when we last updated the website back in 2021, so maybe this is an update in git?

gonuke:
I don't think it's related to changes in git.  I recognize this failure from way back, and I'm not sure how/when the `source` directory ended up in our `master` branch. I don't think it's supposed to be there...  I'm digging through old commits to try to understand (or figure out why I remember this incorrectly...)

abachma2:
Have you had time to look into/think about this any?

gonuke:
> Have you had time to look into/think about this any?

I tried to figure out if we need/expect the `source` directory to be there in the final published version of the website.  My recollection is that we should not need that directory and that it's not supposed to be there.  However, when I look through old commits, that directory has been present for a very long time - I think.

I'm trying to figure out how to confirm this, however, because we force push to `main` when we publish, so if the `source` directory was added recently, it may appear to be long standing in the git logs.

I guess the real question is whether anything on the website links into that directory and whether it would break the website if that directory was gone?

gonuke:
I searched a recent local build of the website and didn't find any links that referred to this `source` folder.

abachma2:
I can always try to remove the source directory from the main branch, then try to build and publish to see if that fixes it. We shuold be able to undo any changes that does.

gonuke:
P.S. We should rename `master` to `main` at some point

munkm:
@abachma2 how about we open another PR deleting the `source` folder from the `master` branch?

gonuke:
> I just tried:
> 
> * checking out master branch
> * removing `source` directory
> * checking out source branch
> * building via `make gh-publish`
>   I still got the fatal error described above, and the source branch reappeared in the master branch.

It looks like this changed in #303 and maybe it was not quite correct...

gonuke:
I may have approved it without understanding carefully enough what was going on

gonuke:
See PRs #346  and #345 to repair this to its originally intended state.

abachma2:
The changes in #345 and #346 fixed what this PR addressed, so this PR is being closed.

PR #339, opened 2024-03-21
Updates to cep3 in preparation for releases 

munkm:
In preparation for https://github.com/cyclus/cyclus/issues/1685 , we've been looking through the release procedure defined in CEP3. The changes in this PR include: 

* removing references to `smbchk.py`. It has been deprecated. 

I will update this PR description with the other changes as we find them.

### DIFF
--- a/source/cep/cep3.rst
+++ b/source/cep/cep3.rst
@@ -29,12 +29,6 @@ stable.)  The projects that are under the release manager's purview are:
 * `Cycamore`_
 * `Cymetric`_
 
-The projects which are not yet under the release managers purview are:
-
-* `Rickshaw`_
-* `CyclusJS`_
-
-
 .. note::
 
     The following full release process only applies to MAJOR and MINOR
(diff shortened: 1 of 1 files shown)


munkm:
@dean-krueger yes we should change everything to `main`. Would you like to do that?

munkm:
CEP3 mentions the `releases` repository. Do we use this widely? It seems like we might not, because we have a conda-forge feedstock but the conda script seems to predate conda-forge.

munkm:
We also need to update cep3 based on the decisions in: https://github.com/cyclus/cyclus/issues/1700

abachma2 (approved):
This looks good to me. Are there any other changes that need to be made to this CEP before merging? @gonuke?

munkm:
I've been going back through this document and have a few questions: 
- Do we want to keep the pyne directions in the maintenance section?
- Do we want to keep the nuclear data directions in the maintenance section?

abachma2:
On a previous call, I mentioned that I think the main place that pyne is used is in the decay solver. Is anyone aware if this is true? And thinking about it some, pyne may also be used in Cycamore, with the FuelFab archetype. That might be worth exploring, since Cycamore has Cyclus as a dependency.

gonuke:
I think PyNE is used in a few places for compositions, nuclide naming and materials.

PR #342, opened 2024-03-21
Update list of contributors

nsryan2:
This PR changes the explicit list of contributors to links to Cyclus, Cymetric, and Cycamore GitHub contributors.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -92,44 +92,17 @@ Contact Us
 
 * `Developer Call times <http://dev-call.fuelcycle.org>`_: Fridays at 3:00PM EST
 
+
 Contributors
 ------------
 
-* Robert Carlsen
+You can find our amazing contributors through their GitHub profiles
+* `Cyclus Contributors <https://github.com/cyclus/cyclus/graphs/contributors>`_
 
-* Denia Djokic
+* `Cycamore Contributors <https://github.com/cyclus/cycamore/graphs/contributors>`_
 
-* Royal Elmore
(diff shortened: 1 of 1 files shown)


abachma2 (changes requested):
Thanks for making this change. One small working change I would suggest.

nuclearkatie (approved):
Thanks for this, @nsryan2! Always nice to replace something that needs to be updated regularly with an automated system/link

gonuke:
I am sure there was a reason we didn't do this before, but can't recall... it might be because these contributor graphs highlight metrics (commits/additions/deletions) that we didn't think were valuable....  But not a reason to disagree with this change.

gonuke:
It's too bad there is not a view that is just a list of contributors... (bonus points for the person who can produce that from git 😀 )

nuclearkatie:
You could use the output of git shortlog, but that would require updates or additional automation 
`git shortlog --summary`

Also, trying it on my machine also results in a number of different people who show up in different ways, perhaps because they used different machines where they set their `git config user.name` in a slightly different way. For example, @abachma2 is on the  list as `abachma2`, `abachmann`, and `Amanda`, and you @gonuke are there as both `Paul P.H. Wilson` and `Paul Wilson`. I'm not sure how/if you can get github usernames.

gonuke:
> Also, trying it on my machine also results in a number of different people who show up in different ways

yeah... that's a problem

gonuke:
@abachma2 - I committed your recommended change - please confirm that you're ok with this and merge.

abachma2 (approved):
Thanks for making this change @gonuke! LGTM.

PR #343, opened 2024-03-21
Remove weekly meeting information

nsryan2:
This PR removes the zoom link and weekly meeting time from contact us section of index.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -90,7 +90,6 @@ Contact Us
 
 * `Developers' mailing list and forum <https://groups.google.com/forum/#!forum/cyclus-dev>`_
 
-* `Developer Call times <http://dev-call.fuelcycle.org>`_: Fridays at 3:00PM EST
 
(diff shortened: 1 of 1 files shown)


source/index.rst
```diff
@@ -90,7 +90,6 @@ Contact Us
 
 * `Developers' mailing list and forum <https://groups.google.com/forum/#!forum/cyclus-dev>`_
 
-* `Developer Call times <http://dev-call.fuelcycle.org>`_: Fridays at 3:00PM EST
```
gonuke, line 93: Someone should make a note that we can continue to have aliases through the name server to facilitate calls.  That is, we can make `dev-call.fuelcycle.org` point to some Zoom call in the future when we restart these

gonuke (approved):
Not sure where to track our little alias trick

gonuke:
Thanks @nsryan2

PR #349, opened 2024-03-25
Publish site via GitHub Actions

### LINKED ISSUE #344: Publish site using Github Actions
Is there any interest to move the publishing process to CI?

What I'm envisioning is:
- "build/test" workflow that generates the site and uploads the html dir as a build artifact (which reviewers can download to serve locally)
- "publish" workflow that runs on merge to source and pushes generated html to master

I'm skipping over lots of detail but I see value in automating this process so there's less responsibility on the contributor to perform steps locally.  Any thoughts? Comments? Objections?

bennibbelink:
Closes #344.   Instead of maintaining a `main` branch to publish the site this will publish the site via a GitHub workflow on any merge to `source`.  Attempting to follow best practices outlined in [GitHub documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site).

### DIFF
--- a/docker/fuelcycle.org-deps/Dockerfile
+++ b/docker/fuelcycle.org-deps/Dockerfile
@@ -1,9 +1,7 @@
 
-FROM cyclus/cymetric
+FROM ghcr.io/cyclus/cymetric_22.04_apt/cymetric:latest
 
 RUN apt-get update
-RUN apt-get install -y --force-yes wget
-RUN apt-get install -y --force-yes python3-pip
-RUN pip3 install sphinx sphinxcontrib-bibtex
-RUN pip3 install cloud-sptheme==1.6
+RUN apt-get install -y --force-yes wget python3-pip
+RUN python -m pip install sphinx sphinxcontrib-bibtex cloud-sptheme==1.6
 
--- a/docker/fuelcycle.org-deps/Dockerfile
+++ b/docker/fuelcycle.org-deps/Dockerfile
@@ -3,5 +3,5 @@ FROM ghcr.io/cyclus/cymetric_22.04_apt/cymetric:latest
 
 RUN apt-get update
 RUN apt-get install -y --force-yes wget python3-pip
-RUN python -m pip install sphinx sphinxcontrib-bibtex cloud-sptheme==1.6
+RUN python -m pip install sphinx sphinxcontrib-bibtex cloud-sptheme
 
--- a/.github/workflows/build_publish.yml
+++ b/.github/workflows/build_publish.yml
@@ -21,16 +21,13 @@ jobs:
         run: |
             make docker-html
 
-      - name: debug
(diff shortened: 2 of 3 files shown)


bennibbelink:
Comment from the linked issue:
> That said, this relies in our custom build docker images having the right config. We probably want that to be part of the workflow somewhere? I'm not sure all the images have Sphinx. I saw that install in your sample build, but I didn't really see where that happened in your sample PR.

I updated `docker/fuelcycle.org-deps/Dockerfile` to build `FROM` our latest `cymetric` image and install Sphinx.  This should ensure that all of our cyclus libraries exist in the docker image.  I also modified the `docker-gh-preview docker-html` targets to `docker build` using the Dockerfile.  Thus in the workflow we just need to call `make docker-html` to build the site.

bennibbelink:
Another advantage of this moving forward is that to publish the site when we make changes to the `cyclus/cyclus` repo we need only run the workflow via `workflow_dispatch` (for example after we update the dependencies in the `cyclus/cyclus` documentation)

docker/fuelcycle.org-deps/Dockerfile
```diff
@@ -1,9 +1,7 @@
...
-FROM cyclus/cymetric
+FROM ghcr.io/cyclus/cymetric_22.04_apt/cymetric:latest
 
 RUN apt-get update
-RUN apt-get install -y --force-yes wget
-RUN apt-get install -y --force-yes python3-pip
-RUN pip3 install sphinx sphinxcontrib-bibtex
-RUN pip3 install cloud-sptheme==1.6
+RUN apt-get install -y --force-yes wget python3-pip
+RUN python -m pip install sphinx sphinxcontrib-bibtex cloud-sptheme==1.6
```
gonuke: Is this pinning to preserve the sidebar menu?  If so, let's add an issue to update our templates for a newer theme version.

gonuke (commented):
LGTM - just one question about the them pinning

docker/fuelcycle.org-deps/Dockerfile
bennibbelink: I am not sure why it is pinned at 1.6, that's just what was there.  I'll make an issue regardless and try to resolve the sidebar complications.

gonuke: If it's not for any reason, maybe let's drop the pin and see what we get?

bennibbelink: I dropped the pin and don't notice any significant changes (if any at all).  If you want you can download the artifact generated by the action and check it out yourself! 😃 

https://github.com/cyclus/cyclus.github.com/actions/runs/8461607773

PR #351, opened 2024-04-01
Add Docker Installation instructions

bennibbelink:
This PR has many small changes like fixing broken links and versions.  The major addition/removal is dropping VM instructions and adding Docker instructions.  The Docker instructions are pretty bare right now, so I am very open to suggestions.

@dean-krueger @nsryan2 I believe you've gone through these docs fairly recently to try to learn how to use Cyclus.  Feel free to chime in!

P.S. if you want to view the version of the site that would result from this PR you can download the artifact that is present on the most recent workflow run triggered by this PR.

I've found lots of other things that we should update throughout the site, this is just some of it.  I didn't want to bloat any single PR.


gonuke:
This looks good to me... Thanks @bennibbelink

Issue #347, opened 2024-03-22
Publishing from master instead of main


abachma2:
With the changes in #345 and #346 it looks like the website is still published from the master branch and not the main branch like it should be. This may be why some of the most recent changes are not captured in the current version of the site.

gonuke:
I think it's something different from that.

1. I didn't rebuild/republish to main
2. There is another PR that includes more comprehensive updates from master -> main

abachma2:
OK. I thought that had already been merged. Merging #341 will close this issue then.

gonuke:
#341 is now merged, but someone needs to rebuild/publish the website to see final updates

abachma2:
I tried to build it, it all runs fine, but I'm not sure it actually includes all of the updates made in #335.

gonuke:
It might be because of errors in the build?  I noticed that `cep0` doesn't render correctly with the new CEP 29 showing up in the directory

abachma2:
Maybe, because the changes from #340 are present.

bennibbelink:
Is this issue still relevant after #349 was merged?

abachma2:
Let me try building the website from the source branch and see what happens.

abachma2:
Based on the terminal output, all the build commands are pointing to the `main` branch. The main point of this issue is resolved, so I'll close the issue.

PR #335, opened 2024-03-14
CEP29

gonuke:
First draft of CEP on packaging (replaces #334 with new CEP number)

PR #353, opened 2024-04-01
Fix CEP 29 render

abachma2:
When CEP 29 was added in PR #335, it was missing a link to the `source/cep/cep29.rst` file so it didn't render correctly on `source/cep/cep0.rst`. This issue was identified as part of #347, and this PR fixes it.


gonuke:
Thanks for this fix @abachma2

PR #354, opened 2024-04-01
Fix Sidebar

### LINKED ISSUE #350: Sidebar is broken
The sidebar no longer shows everything that it should.  This may(?) be related to the Sphinx version and/or the theme we are using.

bennibbelink:
This modifies our custom sidebar template to work with current versions of Sphinx and cloud-sptheme.  Closes #350.


gonuke:
Thanks @bennibbelink - this looks good.  I don't think all the links in the sidebar work for a local build and that's not a new problem.  They rely on URLs that start at the server root.  TL;DR: that's not a reason to hold this PR up

bennibbelink:
I moved it back into draft because the logo still isn't showing up in the sidebar, but I guess that doesn't need to block this PR either.

gonuke:
Yeah - let's take this step and maybe it will help us figure out where the logo went

Issue #337, opened 2024-03-21
General updates to the website


abachma2:
This issue is meant as a list to keep track of various changes on the website, such as updates, typos, redesigns, etc. We will do our best to make sure things get updated in a timely manner.

nsryan2:
From our initial discussion:
* Remove mentions of VM support (people can build their own VM and then install cyclus inside it)
* Remove mentions of cyclus-build-deps conda install mentions
    * There’s a flag you can use to install only dependencies from a condo package
* Update installation minimum packaging
* Find some way to document new releases

abachma2:
In the build process, there is a warning:
```
WARNING: Failed to import cyclus.lazyasd.
Possible hints:
* ModuleNotFoundError: No module named 'cyclus.lazyasd'
* AttributeError: module 'cyclus' has no attribute 'lazyasd'
```
which points to `source/python/lazyad.py`. I'm not sure what this page does, and it can still build without this, so maybe we can get rid of it?

abachma2:
Update the website with new version information.

nuclearkatie:
[Cyclus publications and meetings](Cyclus Publications & Meetings) appears to have last been updated in 2015

nuclearkatie:
- [ ] [mbmore](http://fuelcycle.org/user/index.html#third-party-archetypes) links point to Meghan's personal repo, when they are now managed and updated using [this cnerg repo](https://github.com/cnerg/mbmore)

nuclearkatie:
- [ ] [Projects Using Cyclus](http://fuelcycle.org/basics/projects.html) currently only lists CVT (which doesn't exist anymore)

dean-krueger:
We could add all the little things I listed in this (https://github.com/orgs/cyclus/projects/3) project since we're messing with the website anyway.

nuclearkatie:
Should [Visualization and Analysis](http://fuelcycle.org/user/index.html#visualization-analysis) also include third-party tools? I'm specifically thinking about [Bicyclus](https://github.com/Nuclear-Verification-and-Disarmament/bicyclus) and [Trailmap](github.com/cnerg/trailmap)

abachma2:
> Should [Visualization and Analysis](http://fuelcycle.org/user/index.html#visualization-analysis) also include third-party tools? I'm specifically thinking about [Bicyclus](https://github.com/Nuclear-Verification-and-Disarmament/bicyclus) and [Trailmap](github.com/cnerg/trailmap)

[...]

PR #355, opened 2024-04-01
Update publication lists

abachma2:
This PR updates some of the publications listed on the website. Currently, the only change is to add the `source/cite/citecyc.bib` file to the bibtext_bibfiles list in `source/conf.py` so that the publications about difference Cyclus versions appears. This fixes one of the items identified in #337.

@nuclearkatie also mentioned in #337 an item to update this list. Were there any specific publications you had in mind? The lists on the website seem to be focused on the core of Cyclus, and not as much on applications. Do we have any newer publications that fit this bill? I think expanding this list to publications that just use Cyclus will be burdensome to maintain, but I'm open to ideas.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -52,7 +52,12 @@ extensions = ['sphinx.ext.mathjax', 'sphinx.ext.autodoc',
               'cyclusagent', 'cloud_sptheme.ext.table_styling',
               'sphinxcontrib.bibtex']
 
-bibtex_bibfiles = ['cite/pubs.bib', 'cep/cep-0002-1.bib', 'cep/cep-0004-1.bib', 'cep/cep-0018-1.bib', 'cep/cep-0020-1.bib']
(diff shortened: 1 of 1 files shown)


gonuke:
While you are at it, can you add others.  I found some in the `newsletters` directory

gonuke (changes requested):
While you are at it, can you add others.  I found some in the `newsletters` directory

abachma2:
I really like the idea in #356, so let's not merge this until that is completed, and it can be a part of this PR.

gonuke:
It may take some time to implement #356.  Might be better to go ahead with this until we sort that out properly.

gonuke (approved):
I'm happy to approve this now, and make #356 a separate PR.  This makes everything work immediately until we can sort out how to do #356.

abachma2:
That works for me. Merge away

PR #358, opened 2024-04-02
Update Makefile and README to reflect new publishing procedure

### LINKED ISSUE #357: Update `README` and `Makefile` to reflect the fact that we build with an action
We now build with a GitHub action instead of publishing from `main`.  (@bennibbelink - does that action push to `main` or to some unidentified place in the cloud that we don’t need to know about?)

To avoid confusion, we should update the instructions and perhaps streamline the Makefile to remove things we no longer need (or that no longer have any effect).

bennibbelink:
Closes #357.  We now publish via GitHub Actions so this PR documents this in the README.  It also removes targets from the Makefile that were intended for publishing the site.

The new minimum version numbers are the versions used by CI (our `cymetric:latest` image).  The actual minimum version numbers are likely lower than is listed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,10 +30,7 @@ help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  gh-preview        to build HTML in directory $BUILDDIR for testing"
 	@echo "  gh-revert         to cleanup HTML build in directory $BUILDDIR after testing"
-	@echo "  gh-publish        to build final version and push from source branch to main branch"
-	@echo "  gh-publish-only   to push from source branch to main branch, assuming already built"
 	@echo "  docker-html       to use docker to build HTML in directory $BUILDDIR for testing"
-	@echo "  docker-gh-publish to use docker to build final version and push from source branch to main branch"
 	@echo "  serve             build+serve html files using Python's SimpleHTTPServer"
 	@echo "  serve-only        serve pre-built html files using Python's SimpleHTTPServer"
 	@echo "  dirhtml           to make HTML files named index.html in directories"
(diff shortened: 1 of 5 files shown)


gonuke (approved):
This looks good to me @bennibbelink - thanks.  I'll request @abachma2 review this as well, since she's been active in website development.

docker/README.rst
```diff
@@ -1,9 +1,6 @@
 
 ``fuelcycle.org-deps`` contains a dockerfile with all dependencies for
```
abachma2: Is this container up to date with what's in the README?

abachma2 (commented):
So the website builds upon merging to the `source` branch, so we don't have to maintain a `main` branch. Are these the only places where we need to update the documentation? Did you `grep` for where ever the commands and the main branch are mentioned? If this is all that needs to up updated for documentation then this looks good to me.

docker/README.rst
gonuke: We should probably just remove this directory entirely?  where else is it referenced?

bennibbelink: That subdirectory is where the Dockerfile is located.  I can move the Dockerfile up to the `docker/` directory and drop the `docker/fuelcycle.org-deps/` subdirectory.

bennibbelink:
> Did you grep for where ever the commands and the main branch are mentioned?

Just went through and made a couple more small changes to remove references to the main branch.  I think now I got them all.

docker/README.rst
gonuke: Oops... I forgot that we were actually still using this dockerfile....

Makefile
abachma2, line 1: In l. 13-14 we have variables `GIT_BRANCH` and `GIT_FORK`. Do we still need those?

gh-project.mk
```diff
@@ -4,15 +4,9 @@
...
-GH_UPSTREAM_REPO = upstream
 
 
 # Example for a gh-pages project
 #
 # GH_SOURCE_DIR = doc-src
-# GH_SOURCE_BRANCH = main
-# GH_PUBLISH_BRANCH = gh-pages
-# GH_UPSTREAM_REPO = upstream
+# GH_SOURCE_BRANCH = main
```
abachma2: Do we still need this?

abachma2 (commented):
Hi @bennibbelink Thanks for going through some more of the build. I have a few questions about spots that we may or may not need anymore.

gh-project.mk
bennibbelink: No we do not, thanks for catching that!

Makefile
bennibbelink: These allow you to fetch different versions of `INSTALL.rst` and `DEPENDENCIES.rst`.  Although - I think it makes more sense to have two sets of variables, one set pointing to a cyclus repository and one set pointing to a cycamore repository.  I updated the Makefile accordingly.

abachma2 (approved):
This looks good to me. Thanks for making those changes @bennibbelink

PR #364, opened 2024-04-17
Fix numbering issues in `user/install_binary.rst`

bennibbelink:
This should be considered along with cycamore PR [#595](https://github.com/cyclus/cycamore/pull/595).  I moved some of this information to `DEPENDENCIES.rst` in the cycamore repo so that the numbering looks better.

@abachma2 you're more than welcome to build the site locally with the two PRs, I've attached what I'm seeing locally:
image

### DIFF
--- a/source/user/install_binary.rst
+++ b/source/user/install_binary.rst
@@ -18,6 +18,8 @@ Installing |Cyclus| with Binaries
    :start-after: .. website_include_deb_start
    :end-before: .. website_include_deb_end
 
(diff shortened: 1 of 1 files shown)


abachma2:
This all looks so much cleaner! Thanks for digging into this. I'll try to build this locally at some point to view it all myself.

gonuke:
Thanks @bennibbelink - looks like it's all working

gonuke (approved):

PR #367, opened 2024-04-29
More general updates

abachma2:
This PR addresses some of the issues raised in #337, specifically:
- some of the build warnings from `source/cep/cep27.rst`
- adding a news item 
- adding a missing version release for v1.5.5.
- fixing typos/small issues identified by @dean-krueger 
- removing the `data_explorer.rst` from the user tutorial
- removing mention of cloud for running cyclus from user tutorial.


source/previous/v1.5.5.rst
```diff
@@ -0,0 +1,57 @@
+.. _1.5.5:
+
```
bennibbelink: In our meeting on Monday we discussed modifying the "Cyclus Previous Versions" page to simply reference the GitHub releases.  The releases already have the changes in the descriptions so it probably isn't necessary to copy/paste that information into the site as well.  Do you have any thoughts about this?

abachma2: So something like pointing to [this page](https://github.com/cyclus/cyclus/releases/tag/1.5.5) for the 1.5.5 release? Instead of having separate web pages for the releases?

bennibbelink: Not sure if we even need to link each release... we could say that all the releases can be found at [this page](https://github.com/cyclus/cyclus/releases) so we don't need to manually copy/paste info from Github to the site with each release.  I might have misunderstood the group consensus though - @gonuke what did you have in mind?

EDIT: Thats exactly what you did sorry :) I should've looked closer at your changes before commenting

bennibbelink:
Not sure if you saw my [PR](https://github.com/abachma2/cyclus.github.com/pull/1) to your branch.  This small change could be thrown into this PR.

source/previous/index.rst
```diff
@@ -5,21 +5,10 @@
...
-    v1.2
-    v1.1
-    v1.0
-    v0.4
-    v0.3
-    v0.2
+Release notes and changelogs for previous versions of |Cyclus| are 
+available through the `project GitHub <https://github.com/cyclus/cyclus/releases>`_.
+These pages also contain zips for the source code of each version. These zip 
+packages are also available below.
```
bennibbelink, line 11: Do we need to maintain this table of zips/tarballs still?

source/news/fy24neup.rst
```diff
@@ -0,0 +1,21 @@
...
+is lead by:
+
+* Dr. Madicken Munk (UIUC, PI)
+* Prof. Paul P.H. Wilson, Ken Petersen (UW-Madison)
+* Dr. Eva Davidon, Jin Whan Bae (ORNL)
+
+
+.. _NEUP: http://www.neup.gov
+
+.. _MARKAL: http://www.iea-etsap.org/web/Markal.asp
```
bennibbelink:
```suggestion

```

source/news/fy24neup.rst
```diff
@@ -0,0 +1,21 @@
...
+
+.. _GCAM: http://www.globalchange.umd.edu/models/gcam/
```
bennibbelink:
```suggestion
```

source/user/tutorial/index.rst
```diff
@@ -30,7 +30,6 @@ Exercise #1
     add_commod_recipe
     add_proto
     add_reg_inst
-    launch_cyclus
```
bennibbelink, line 33: I see this removes "Using Cyclus in the Cloud".  There is another reference to the cloud service in `user/tutorial/install_launch_cyclus` that should probably also be removed.

bennibbelink (changes requested):

source/previous/index.rst
gonuke: While I agree that it would be great to avoid maintaining the table, it's nice to get a compact view of releases/release dates.... wonder if can automate that in a future update (i.e. make an issue)

source/user/tutorial/data_explorer.rst
gonuke, line 1: Why are we getting rid of this?

gonuke (changes requested):
Not sure why we're deleting the data explorer page?

source/user/tutorial/data_explorer.rst
abachma2: 1. It's not actually called anywhere, like no other page links to it
2. It uses a python package (`analaysis.py`) that I don't know where people are supposed to get access to it. 
3. We link to the Cymetric GitHub in one of the pages for how to use that tool for analysis
4. We are missing a bunch of figures that are supposed to go on this page, and I think it would be nontrivial to recreate them, especially since I don't have the `analysis.py` to replicate the procedure.

source/previous/index.rst
abachma2: All of the zips are on the github tags page. Maybe link to those?

source/user/tutorial/index.rst
abachma2: I'll make sure that gets removed. This cloud service isn't working, so we shouldn't get users' hopes up that they don't have to install locally.

abachma2:
One note about removing the release notes from the website: Some of the Release notes with the tags on GitHub point to those pages ([like v1.3](https://github.com/cyclus/cyclus/releases/tag/1.3.0)). Should we just copy the contents of the web pages to these notes? To make sure we don't lose anything and it's all in the same place?

source/user/tutorial/data_explorer.rst
gonuke: Thanks, I it didn't appear to be so immediately out of date on my scan during review, but that all makes sense.

abachma2: Fair question. And I think I included some of this information in the commit message when I deleted it, so hopefully it helps to answer future questions.

gonuke:
> One note about removing the release notes from the website: Some of the Release notes with the tags on GitHub point to those pages ([like v1.3](https://github.com/cyclus/cyclus/releases/tag/1.3.0)). Should we just copy the contents of the web pages to these notes? To make sure we don't lose anything and it's all in the same place?

It looks like v1.3 may have been the only one and I just copy/pasted the content to the release info on each of Cyclus/Cycamore/Cymetric

gonuke (approved):

bennibbelink (approved):

PR #370, opened 2024-05-18
Deprecate `ResourceBuff` 

bennibbelink:
- Updates the toolkit documentation to reflect the deprecation of `ResourceBuff`
- Updates Dockerfile to use `python3` instead of version agnostic `python`

### DIFF
--- a/source/arche/toolkit.rst
+++ b/source/arche/toolkit.rst
@@ -6,10 +6,6 @@ Currently in an experimental state, the |cyclus| toolkit aims to provide
 functionality relevant to a variety of :term:`archetypes <archetype>` that do
 not belong strictly in the :term:`cyclus kernel`.
 
-.. note::
-
-    ResourceBuff has been deprecated in favor of ``ResBuf``.
-
 ResBuf
 ++++++++++++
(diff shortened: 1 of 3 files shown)


source/arche/toolkit.rst
```diff
@@ -6,66 +6,9 @@ Currently in an experimental state, the |cyclus| toolkit aims to provide
...
-
-        # Option 3: you can reference another annotated member variable's value
-        # as the capacity
-        buf_cap = ts.Double()
-        mybuf = ts.ResourceBuffInv(capacity='buf_cap')
-
-You can read the `ResourceBuff API documentation
-<http://fuelcycle.org/cyclus/api/classcyclus_1_1toolkit_1_1ResourceBuff.html>`_ for
-more details on how to use the buffer.
+    ResourceBuff has been deprecated in favor of ``ResBuf``.
```
gonuke: I wonder where it is best to include this note?  Maybe at the end of the `ResBuf` preamble?

gonuke (changes requested):
Just wondering about the placement of the deprecation note.

source/arche/toolkit.rst
bennibbelink: Here?

image

gonuke (approved):

PR #381, opened 2024-08-03
Add docs for new constants

### LINKED ISSUE #380: Add instructions for using constants in archetype development
https://github.com/cyclus/cyclus/pull/1757 and https://github.com/cyclus/cycamore/pull/606 added constants for large values.  They can be used to develop archetypes (see cycamore for examples).  This should be documented in the site with instructions/examples for how to use these constants.

bennibbelink:
Closes #380.

I also modified the dockerfile to no longer `mamba update -y --all` because it was updating libraries that cyclus depends on (specifically `libicudata.so` could not be found), rendering cyclus unusable.  I'm not sure if we will run into trouble later on of mamba not being able to find the sphinx packages required for building the site.  I think that the update command might be a remnant of using an apt build (needing to update package indexes in order to find the sphinx packages), and I'm not sure that it's necessary with conda/mamba.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,3 @@
 FROM ghcr.io/cyclus/cymetric_22.04_conda/cymetric:latest
 
-RUN mamba update -y --all
 RUN mamba install -y sphinx sphinxcontrib-bibtex cloud_sptheme
--- a/source/arche/constants.rst
+++ b/source/arche/constants.rst
@@ -50,6 +50,25 @@ the ``cyclus`` package is in your Python path:
   my_small_value = CY_NEAR_ZERO
   ...
 
+This holds true for |Cyclus| input files: 
+
+.. code-block:: python
+
+  from cyclus.system import CY_LARGE_DOUBLE
+  SIMULATION = { 'simulation': 
+    ...
+    'facility': [
+      { 'config': 
+        { 'Source': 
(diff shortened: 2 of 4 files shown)


bennibbelink:
@gonuke feel free to tag some active archetype developers.  I would love to hear their feedback on the docs as well but am not sure who all to bother :smile: nuclearkatie? abachma2?

nuclearkatie:
Sorry if this is a basic question but I haven't been following this saga closely. Are the new variables able to be used in prototype declarations in input files, or only in core cyclus and archetype code bases? Like, can you declare a source with 

```
<facility>
  <name>ExSource</name>
  <config>
    <Source
      <outcommod>commod</outcommod>
      <inventory_size>CY_LARGE_DOUBLE</inventory_size>
    </Source>
  </config
</facility>
```
in an input file and have it process correctly? I think either way, the docs should be clear about whether you can do this

bennibbelink:
@nuclearkatie good question. They can be used in input files written in Python, but not XML.  The latest commit clarifies this in the docs

katyhuff (approved):
This looks reasonable to me. I suspect the mamba issue you've fixed here might solve some installation issues one of our ARFC undergrads may be experiencing. I have no recommended edits.

bennibbelink:
If there are no further requests for changes, can someone merge this please?

PR #385, opened 2024-09-13
Rewrite DbTypes table

### LINKED ISSUE #384: Database Types table needs fixing
https://fuelcycle.org/arche/dbtypes.html# See this link, it appears that the table exists but there are no rows

bennibbelink:
Closes #384.

I wasn't able to get the old version to render correctly so I'm not 100% sure what it was supposed to look like, but I think this working version looks alright.  It may(?) have been a pivot table previously, and if we want that functionality back it wouldn't be hard to implement (https://pivottable.js.org/examples/), but for now it is just a pretty basic table (with search/sort functionality).

Have a look and let me know what you think.


gonuke:
I don't see it in the artifact from testing... do I need to build locally to see it?

bennibbelink:
Are you sure you're serving the correct build artifact? This is what I see from the artifact in this workflow run https://github.com/cyclus/cyclus.github.com/actions/runs/10840934866?pr=385

Screenshot 2024-09-12 at 9 06 10 PM

gonuke:
Strange... I get this....
Screenshot 2024-09-12 at 20 16 46

but when I look at the source, I see all the right source there to render the table...??

gonuke:
Oh... it looks like you're serving it on localhost - I'm just loading the file...

gonuke:
found a quick server: `python3 -m http.server` and it worked!

bennibbelink:
> Oh... it looks like you're serving it on localhost - I'm just loading the file...

Yep I always preview it using `python3 -m http.server`

gonuke (approved):
Thanks @bennibbelink - I don't think we were actually using the pivot table capabilities and this has the capabilities that matter

PR #387, opened 2024-09-19
Build api docs when building site

### LINKED ISSUE #1732: State of doxygen docs
I am uncertain as to the state of the `doc/` directory.  I built the docs locally and it appears to produce what is currently at http://fuelcycle.org/cyclus/api/.  According to the [style_guide ](https://github.com/cyclus/cyclus.github.com/blob/source/source/kernel/style_guide.rst) this page is updated nightly, which I am skeptical of :smile:.  

What is the purpose of these doxygen docs - should they be rid of entirely or do we still have a need for them (in which case they need an overhaul)? 

These questions arose from the existence of `doxygen` as a dependency when working on #1710.

bennibbelink:
Moves the site building action into the docker build, and adds some steps to build the Cyclus and Cycamore api docs in the dockerfile too.

Closes https://github.com/cyclus/cyclus/issues/1732

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,8 +74,10 @@ gh-preview html:
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)."
 
 docker-gh-preview docker-html:
-	docker build -f docker/Dockerfile -t site-deps .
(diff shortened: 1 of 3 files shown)


gonuke (approved):
This looks good to me... thanks @bennibbelink

PR #410, opened 2025-01-17
Readme build instructions

### LINKED ISSUE #397: cyclus.github.com README is unclear on how to actually serve test versions of the website

dean-krueger:
Previous versions of the README left users (me when I was trying to do this for the first time) high and dry after telling them to run `make docker-gh-preview`. It was not at all clear that to actually view the changes made to the website it needed to be served and visited at localhost. This version of the README gives users (especially new ones) more detailed instructions on exactly what they need to do after they make changes to the website so that they can check their work.

Closes #397

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,7 +82,28 @@ Best practice workflow for contributing to site changes
    ``make docker-gh-preview``
 
    to build the website inside a docker container with all the correct
-   dependencies and configuration taken care of automagically.
+   dependencies and configuration taken care of automagically. Once this
+   is done, navigate into the gh-build directory and serve the website.
+
+   ``cd gh-build``
+   ``python3 -m http:server <port number>``
+
+   the most common port number for things like this is 8000, or 8080,
+   and as such choosing one of these is recommended:
+
+   ``python3 -m http:server 8000``
+
+   If you are building the website on a computer which you are 
+   connected to remotely, it will then be necessary to create a tunnel
+   between your local and remote machines. This can be done with:
+
+   ``ssh -L <port number>:localhost:<port number> username@remote``
+
+   So if using port 8000 as in the example above, and connecting to a
+   remote machine which you had listed in your config file as "remote-machine"
+   the command would look something like this:
+
+   ``ssh -L 8000:localhost:8000 username@remote-machine``
 
 6. Repeat steps 4-5 until satisfied.
 
--- a/README.rst
+++ b/README.rst
@@ -85,8 +85,12 @@ Best practice workflow for contributing to site changes
    dependencies and configuration taken care of automagically. Once this
    is done, navigate into the gh-build directory and serve the website.
 
-   ``cd gh-build``
-   ``python3 -m http:server <port number>``
+
+   ::
+      
+      cd gh-build
+      python3 -m http:server <port number>
+   
 
    the most common port number for things like this is 8000, or 8080,
    and as such choosing one of these is recommended:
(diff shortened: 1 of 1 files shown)


dean-krueger:
should be `python3 -m http.server <port number>`. Oops. I'll change that momentarily.

README.rst
```diff
@@ -82,7 +82,41 @@ Best practice workflow for contributing to site changes
...
+      cd gh-build
+      python3 -m http:server <port number>
+   
+
+   the most common port number for things like this is 8000, or 8080,
+   and as such choosing one of these is recommended:
+
+   ``python3 -m http:server 8000``
+
+   If you are building the website on a computer which you are
```
gonuke, line 100: I think there are two issues here.  In general, you could access the remote machine directly as:
```
https://remote.machine:8000
```

But if that remote machine is behind a firewall then you will need port-forwarding as well.

gonuke (changes requested):
A comment about port-forwarding and when it's necessary.

Also, does it make sense to describe how to serve it from inside a docker container and access it?  Or do we assume/instruct people to bind mount everything and then launch the server from outside the docker container?

README.rst
dean-krueger: Huh, I didn't know you could access remote machines like that. You're saying that for, say, non-UW machines you'd be able to just connect to your remote machine's IP: <port number> to see the page (assuming there wasn't some other firewall)? I've never done that method before, and so I'm not totally certain how it works. However, especially for our workflow here at UW the method I outlined gets the job done. This leads me to wonder how big the Venn diagram intersection of "people who want to contribute to the website who aren't associated with UW" and "people who want to build the website on a remote machine and access it through a local machine" is. Let me know if you think it's worth it for me to write a little section about what to do in that case. I THINK what I wrote will work for all cases, so maybe it's best to just leave that as the default instruction (I'm not aware of any big issue with doing `ssh -L 8000:localhost:8000 username@remote.machine`, but I'm also not an expert).

dean-krueger:
> should be `python3 -m http.server <port number>`. Oops. I'll change that momentarily.

Fixed

README.rst
gonuke: I think it may make sense to have a separate section of the readme about serving the preview, and pointing to that section from here.

It could have subsections:
* serving with http.server
* accessing directly: locally or remotely w/o tunnel
* accessing through tunnel

dean-krueger: not sure if this ever got addressed, but putting it back on my active to-do list as I clean up some GitHub notifications. Hopefully I can get some movement on this soon and we can wrap this up.

gonuke: Looking again, probably those bullets are the in the order of priority, falling off rapidly

gonuke: Oops - this thread is out of date maybe

gonuke (approved):
LGTM - thanks @dean-krueger

gonuke:
Oh - needs a rebase or similar

dean-krueger:
I hit the "Update Branch --> Rebase" button here on GitHub and it looks like it did a bunch of stuff automatically for me. Not entirely sure if that was good or not, so I'm going to take another quick look. It might be worth the same from @gonuke before we go and hit any buttons (I'll probably not hit that button again).

dean-krueger:
It looks like this should be fine? I guess worst case scenario it messes up the website and then I spent the weekend fixing it.

gonuke:
Thanks @dean-krueger

Issue #412, opened 2025-04-10
https://cyrun.fuelcycle.org is down


nsryan2:
In #337, @dean-krueger identified that https://cyrun.fuelcycle.org is down. This appears to still be the case. 

Edit for more information:

After talking with Dean on Slack, he identified that it is linked on https://fuelcycle.org/user/tutorial/launch_cyclus.html.

dean-krueger:
I could be wrong, but I think this was the cloud service that used to be used to run CYCLUS on the web. I'm not certain if there are plans to revive this or not, so I'm not certain what should be done about this.

gonuke:
This was a web service that used to run on a commercial cloud server.  For some time, that was provided to the Cyclus community for free, but I think that expired.  We could continue to run something for probably only a few $$ per month.  An alternative would be to develop an on-demand service that only fired up a resource when requested.  Many years ago, I thought of using [AWS Lambda](https://aws.amazon.com/pm/lambda) for this but that was just conceptual.

abachma2:
I see two main ways to close this issue:
1. Remove the link from the website
2. Getting this service up and running. 

Is there a preference? @gonuke @dean-krueger @nsryan2

gonuke:
I definitely prefer removing it for now - we can make an issue to replace it with something, but hopefully we can make things easy enough to use that we don't need to...

gonuke:
Closed by #417

PR #417, opened 2025-05-23
Remove file for running on the cloud

abachma2:
This PR removes the `source/user/tutorial/launch_cyclus.rst` from the tutorial section of the website. This page is not called anywhere in the tutorial (I think it used to be), and it contains information related to the server on `cyrun.fuelcycle.org` that is no longer available. 

This PR addresses #412.


nsryan2 (approved):
LGTM! Thanks for addressing this extra file. I'll give others a little more time to review, but I'm happy to merge now as this seems pretty cut-and-dry

gonuke:
Was gonna check the autogenerated build, but I'm too late :) - I think this is straightforward.  Thanks @abachma2

PR #418, opened 2025-05-30
Remove long-deprecated files.

gonuke:
It seems like these files were renamed to remove them from the documentation build because they were no longer relevant.  They no longer build as part of our documentation and haven't done so for over a decade, perhaps longer. This commit removes them permanently.

Issue #366, opened 2024-04-29
C++ Tutorial "Hello World" has incorrect install instructions


abachma2:
in the C++ Hello World tutorial it tells you to clone cyclus/cycstub but the cycstub github page has different instructions and says that the method described on the tutorial page is "Depreciated".

abachma2:
@dean-krueger identified this as a needed update to the website. The Cycstub repo says that upon the next release (i.e., 1.6 at this point) it will be removed from Github. We should then think about this one. I haven't done much c++ archetype development, but this did seem like a handy way to help new developer. Having said that, it probably needs some updates should we decide to keep this. 

I welcome discussion on this one.

gonuke:
This repo is replaced by a CLI that lives in cyclus that accomplishes the same thing.  Anyone with cyclus installed can run
``` bash
$ cycstub --type facility mypath:mylib:MyFacility
```
and get the same thing.  A documentation change should take care of it.

abachma2:
That's super useful! I'll give it a try later to make sure it still works well and then try to update the documentation for this.

abachma2:
OK. So quickly playing around with it, it does work (I think). It created a new cyclus input file to run the archetype with a NullRegion and NullInst (did --type facility). Not quite sure if it creates the archetype files that you're then supposed to modify to build your archetype.

gonuke:
I think this was resolved in #419

PR #419, opened 2025-05-30
Wipe cycstub repo

### LINKED ISSUE #197: Scour for cycstub references
* Remove cycstub repo reference from Developers tutorial

gonuke:
This removes references to the cycstub repo and instead replaces it with reference to the cycstub utility.  This leaves one file untouched because it is slated to be removed in #418 .

`cmake.rst` still needs a closer review, but this could be left to a different PR??

Fixes #197 and #366

### DIFF
--- a/source/arche/cmake.rst
+++ b/source/arche/cmake.rst
@@ -3,7 +3,7 @@
 Building Modules with CMake
 ===========================
 
-If you haven't follow the initial example in :ref:`hello_world_cpp`, you should use
+If you haven't followed the initial example in :ref:`hello_world_cpp`, you should use
 ``cycstub`` to generate a directory with the "stubs" to build a new archetype.
 (see :ref:`hello_world_cpp` for further instructions).
 
--- a/source/arche/cmake.rst
+++ b/source/arche/cmake.rst
@@ -3,13 +3,11 @@
 Building Modules with CMake
 ===========================
 
-If you haven't follow the initial example in :ref:`hello_world_cpp`, you should get
-the `Cycstub repo <https://github.com/cyclus/cycstub>`_ by either `cloning the
-repository <https://github.com/cyclus/cycstub.git>`_ or by `downloading the zip
-file <https://github.com/cyclus/cycstub/archive/develop.zip>`_ (see
-:ref:`hello_world_cpp` for further instructions).
+If you haven't follow the initial example in :ref:`hello_world_cpp`, you should use
+``cycstub`` to generate a directory with the "stubs" to build a new archetype.
+(see :ref:`hello_world_cpp` for further instructions).
 
-The Cycstub repo provides a number of critical tools for building your own
+The ``cycstub`` utility provides a number of critical tools for building your own
 module:
 
 * ``install.py``: a basic installation script
--- a/source/arche/cmake.rst
+++ b/source/arche/cmake.rst
@@ -95,12 +95,44 @@ listing:
 
 Examples
 --------
-    
+
+Module Installation
++++++++++++++++++++
+
+The "Hello World" example only allows you to add multiple archetypes to the same
+module/library, using the `USE_CYCLUS` and `INSTALL_CYCLUS_MODULE` macros.  If
+you call `cycstub` multiple times in the same folder, it requires you to specify
+the same path and library name for each new archetype.  This approach reduces
+the number of libraries and executables that are added to a system, perhaps
+unnecessarily.
+
+The resulting `src/CMakeLists.txt` file will includes lines like this:
+
+.. literalinclude:: module-cmake
+
+This will generate:
+
+* a single shared object library in ``$CYCLUS_INSTALL_PREFIX/lib/cyclus/tutorial``
+  named ``libTutorialLibrary.so`` (\*nix) or ``libTutorialLibrary.dylib`` (mac)
+
+* a single unit test executable in ``$CYCLUS_INSTALL_PREFIX/bin`` named
+  ``TutorialLibrary_unit_tests``
+
+where both incorporate the ``TutorialFacility``, ``TutorialInstitution``, and
+``TutorialRegion`` :term:`archetypes <archetype>`.
+
+
 Standalone Installation
 +++++++++++++++++++++++
 
-Through the :ref:`hello_world` example, three standalone modules are installed
-using a ``src/CMakeLists.txt`` file that looks something like
+There may be cases where one prefers to have separate libraries and executables
+for each archetype.  This is not currently supported by `cycstub` in a direct
+way. However, you can modify the above module approach by editing
+``src/CMakeLists.txt`` and replacing:
+
+.. literalinclude:: module-cmake
+
+with 
 
 .. literalinclude:: standalone-cmake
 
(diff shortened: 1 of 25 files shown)


source/arche/cmake.rst
```diff
@@ -3,13 +3,11 @@
 Building Modules with CMake
 ===========================
 
-If you haven't follow the initial example in :ref:`hello_world_cpp`, you should get
-the `Cycstub repo <https://github.com/cyclus/cycstub>`_ by either `cloning the
-repository <https://github.com/cyclus/cycstub.git>`_ or by `downloading the zip
-file <https://github.com/cyclus/cycstub/archive/develop.zip>`_ (see
-:ref:`hello_world_cpp` for further instructions).
+If you haven't follow the initial example in :ref:`hello_world_cpp`, you should use
```
abachma2:
```suggestion
If you haven't followed the initial example in :ref:`hello_world_cpp`, you should use
```

source/arche/hello_world_cpp.rst
```diff
@@ -4,48 +4,32 @@ Hello, Cyclus! [C++]
...
 
-First, you need to get the ``cycstub`` code.  Cycstub is a skeleton code base
+First, you need to use ``cycstub`` to generate a skeleton code base
 that you can use to quick-start new |cyclus| module development projects.
-You can grab cycstub either by using git to
-`clone the repository <https://github.com/cyclus/cycstub.git>`_ or by
-`downloading the zip file <https://github.com/cyclus/cycstub/archive/develop.zip>`_.
-Let's put this code in a ``tutorial`` directory and go into it.
+``cycstub`` is installed with the main ``cyclus`` kernel and should be
+available whereever ``cyclus`` is available.
```
abachma2:
```suggestion
available wherever ``cyclus`` is available.
```

source/arche/hello_world_cpp.rst
```diff
@@ -4,48 +4,32 @@ Hello, Cyclus! [C++]
...
+that should already be complete for building a new archetype.  Without any changes, this archetype
+will not do anything.
 
 ------------
 
 Let's now change the behavior of the TutorialFacility's ``Tick()`` &
 ``Tock()`` member functions to print "Hello" and "World" respectively.  To do
 this, please open up the :file:`src/tutorial_facility.cc` file in your
-favorite text editor (vim, emacs, gedit, `notepad++ <http://exofrills.org>`_).
+favorite text editor (vim, emacs, gedit, `VS Code <https://code.visualstudio.com/>`_).
```
abachma2:
```suggestion
favorite text editor (e.g., vim, emacs, gedit, `VS Code <https://code.visualstudio.com/>`_).
```

source/arche/tour_cpp.rst
```diff
@@ -2,15 +2,17 @@ A Tour of Cycstub
...
+stubs for a Facility, Institution or Region, respectively, is:
+
+.. code-block:: bash
 
-* ``StubFacility``
-* ``StubInstitution``
-* ``StubRegion``
+  cycstub --type facility :stublibrary:StubFacility
+  cycstub --type institution :stublibrary:StubInstitution
+  cycstub --type region :stublibrary:StubRegion
```
abachma2, line 15: Do we want these to have CamelCase in the library name?

abachma2 (commented):
This generally looks good to me. I recommend some of the capitalization is consistent with [1856 in Cyclus](https://github.com/cyclus/cyclus/pull/1856#pullrequestreview-2882373577).

source/arche/tour_cpp.rst
gonuke: I did it this way before I committed to fixing the CamelCase problem. In the fixed version, all combinations will work successfully, so these don't need to change.

source/arche/hello_world_cpp.rst
```diff
@@ -4,48 +4,32 @@ Hello, Cyclus! [C++]
 ====================
 This pages walks you through a very simple hello world example using
 |cyclus| agents.  First make sure that you have the dependencies installed,
-namely |Cyclus|, CMake, and a recent version of Python (2.7 or 3.3+).
+namely |Cyclus|, CMake, and a recent version of Python (3.3+).
```
gonuke: I'm not sure what our minimum version is these days????

gonuke:
I reverted this to draft because a little more work needs to be done to make it all consistent and useful.

source/arche/hello_world_cpp.rst
abachma2: My guess would be python 3.9, but I can look around.

gonuke:
I think this is probably where this needs to be to be consistent now.

source/arche/hello_world_cpp.rst
abachma2: Going off of the dependencies for installing cyclus via conda, it looks like 3.8 is the lowest version allowed. I can build cyclus with 3.8 locally, but the unit tests won't pass unless you have python 3.11+, as of commit 4d1eeef5

dean-krueger (commented):
The only two comments I have are the same PascalCase vs camelCase comment I made in the other PR #1856, and that (and I'm not sure if this really needs to be fixed), but a naive version of me would probably try to type the command `tutorial $ cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution` instead of the command `cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution` in the tutorial folder. Perhaps if you copy the command from the markdown file it is smart enough to not give you that part, but just a thought! Otherwise this looks good to me.

gonuke:
> but a naive version of me would probably try to type the command `tutorial $ cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution` instead of the command `cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution` in the tutorial folder

I think we've often wondered whether it makes sense to include the prompt in the samples.  I'm torn on this.  The only real benefit is that you can see the directory change and the user gets some clear feedback on which directory they should be in to run the command.  The downside is that it looks like part of the command... we could try a different prompt separator... maybe:
```
(/home/tutorial) |> command here
```

dean-krueger:
I'm not sure the separator is really the problem, since ultimately this "issue" is with people who aren't that familiar with using the terminal. I wonder how other people handle this when writing documentation. I guess my personal preference is to change:

"""
You can add additional agents to this module by calling `cycstub` multiple times:

```
    tutorial $ python rename.py tutorial
    tutorial $ cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution
    tutorial $ cycstub --type region tutorial:TutorialLibrary:TutorialRegion
```
"""

Into something like:

"""
You can add additional agents to this module by calling `cycstub` multiple times.

 From the "tutorial" directory, run: 
```
    python rename.py tutorial
    cycstub --type inst tutorial:TutorialLibrary:TutorialInstitution
    cycstub --type region tutorial:TutorialLibrary:TutorialRegion
```
"""

But perhaps that's too verbose/much overhead/easy to miss? Ultimately I don't think it matters much (except that actually this makes the commands really easy to copy form the markdown file and paste in your terminal since you don't have to remove the `tutorial $` bit from every line), but wanted to point this out.

gonuke:
I just updated all the `code-block`s to be type `console` instead of type `bash` to hopefully get syntax highlighting to help with this

gonuke:
All the bash code blocks now have sytnax highlighting that requires the prompt to be indicated with only a single `$` (there may be other options, but indicating the directory name is definitely not allowed).

dean-krueger (approved):
Thanks @gonuke for making that change. This looks good to me now. I'll leave it unmerged until Wednesday morning to give others a chance to look at it on Monday if they want (I know @abachma2 mentioned wanting some time after TWOFCS to check things). If others get around to it before then, and approve, feel free to merge.

abachma2 (approved):
Thanks @gonuke. I am approving this PR. I have one comment to follow up on the minimum python version. If we don't have any more conversation about that then I will merge.

abachma2 (commented):
LGTM. Merging now.

PR #444, opened 2026-01-27
Reducing the verbosity of the tutorial

### LINKED ISSUE #442: [Task] Make tutorials less verbose
# What is this task about?
This issue address how verbose the tutorials are with XML snippets. Sections often give the general form as a template, repeat the template in the activity, and then give the template filled in with the correct values. This task suggests removing the repeated instances of the XML snippets, leaving only the general form for each section. For example, for prototypes, the general form for each archetype would be the only thing given.

# Why is this task needed?
The redundant information hampers the flow of the tutorials and could confuse new users.

# Relevant resources or references

# How can this issue be closed?
This issue can be closed when the repeated instances of XML snippets have been removed, leaving only the general form for each section.

ConnorWelch-OSU:
# Summary of Changes
This PR address the verbosity of the tutorial identified in #442. To reduce verbosity, repeated examples of XML blocks were removed, leaving only a template for each section of an input file. Additionally, minor changes to the text around where XML blocks were removed has been edited to fit the new context that it exists in.

## Design Notes
I did not remove the XML blocks that users can use to check that their input files are correct that are at the end of some pages. There is a discussion to be had on whether these are something that we want to keep or not.

Check the box if your change does not break any of the following:
- [x] API for Cyclus modules
- [x] Input files
- [x] Output tables for post processing tools

# Related CEPs and Issues

Fixes #442 

# Associated Developers

# Testing and Validation

- [x] I have read the [Contributing to Cyclus](https://fuelcycle.org/kernel/contributing_to_cyclus.html) guide
- [ ] I have compiled and run the code locally
- [ ] I have added or updated relevant tests
- [ ] I have added documentation for new or changed features
- [ ] This code follows the style guide
- [ ] I have updated the changelog

Reviewers, please refer to the Cyclus [Guide for Reviewers](https://fuelcycle.org/kernel/pr_review.html).


source/user/tutorial/sim_parm.rst
```diff
@@ -102,42 +102,6 @@ with the variables listed in the table below in your favorite text editor.
...
-        <startyear>2018</startyear>
-        <decay>never</decay>
-      </control>
-
-
-    </simulation>
-
 **Note**: There are two blank lines between the end of the control section and 
 end of the simulation section. This section of the simulation block will hold 
 the rest of the simulation parameter blocks (commodities, facilities, regions,
```
abachma2, line 107: Do we still need this if we don't have the completed control block anymore?

abachma2 (commented):
Thanks for doing this @ConnorWelch-OSU. I think this PR accomplishes the goal of removing some verbosity from the tutorial. 

In the `add_arche`, `add_commod_recipe` and `sim_parm` files, it looks like you took out the completed blocks for users to compare against, but kept those in `add_proto` and `add_reg_inst`. Any reason for that inconsistency? I think having the completed blocks in the tutorial before the final input file has value.

source/user/tutorial/sim_parm.rst
connor-welch: My thinking in leaving it was that the template that appears earlier in the page does contain the two blank lines. It might make more sense to have it further up the page in the section that describes the different parts of the template. I can make that change if its what we want.

abachma2: Please add in those sections. I think they are valuable and make things more consistent.

ConnorWelch-OSU:
> In the `add_arche`, `add_commod_recipe` and `sim_parm` files, it looks like you took out the completed blocks for users to compare against, but kept those in `add_proto` and `add_reg_inst`. Any reason for that inconsistency?

I agree that having completed blocks has value. My opinion is that the completed block should appear in a "Check" section at the end of each page, and that's how they are in `add_proto` and `add_reg_inst`. I can add completed blocks back to the ends of the other pages in a "Check" section, if everyone thinks that's a good idea.

source/user/tutorial/sim_parm.rst
connor-welch: Thanks for the feedback. I have added them in.

abachma2 (approved):
Thanks for making that change @ConnorWelch-OSU. I am approving this PR. I'll leave this open for another few days, in case anyone else wants to look at this. Feel free to ping me to merge this if you think enough time has passed and I haven't merged this yet.

source/user/tutorial/add_reg_inst.rst
```diff
@@ -365,60 +338,11 @@ States Nuclear`` holds the ``UraniumMine``, ``EnrichmentPlant``, and ``NuclearRe
...
-      <config>
-        <NullInst/>
-      </config>
-    </institution>
-
-
-
-
-  </region>
+Let's build the ``ReactorUtility`` institution. This institution has one ``1178MWe ReactorPlant Unit 1`` prototype.
```
meg-krieg: Is this redundant compared to the above lines still under Activity?

source/user/tutorial/add_reg_inst.rst
meg-krieg: Should these be lib5/arch5 and lib6/arch6?

source/user/tutorial/add_reg_inst.rst
```diff
@@ -317,9 +286,13 @@ template.
 | ``region_name``       | ``USA``                     | N/A      |
 +-----------------------+-----------------------------+----------+
 
+Once complete, append this facility under the Sink prototype of your input file.
+
 Check: Complete Region block
 +++++++++++++++++++++++++++++++++++
 
+The region section og your input file should now look like:
```
meg-krieg:
```suggestion
The region section of your input file should now look like:
```

source/user/tutorial/sim_parm.rst
meg-krieg: I think the added completed control block is good for the beginner as this page is their first introduction to seeing completed cyclus input!

source/user/tutorial/add_proto.rst
meg-krieg, line 432: Maybe this comment isn't for this PR, but is there a better home for this footnote (and the other repeats of this footnote)? Would go with the "reduce" verbosity goal of this PR.

meg-krieg (changes requested):
Most of my comments are cosmetic except for 1 typo. 

I do still think the Add Region and Institution activity could be even less verbose. There are a lot of sequences of xml blocks to get to the general template... I am not sure how that reduction would look though.

source/user/tutorial/add_proto.rst
connor-welch: I agree that there would be value in stating the information from the footnote early on in the tutorial, but I decided to leave it for this PR so it can be discussed as a separate issue.

source/user/tutorial/add_reg_inst.rst
connor-welch: Yes. I will make those changes.

source/user/tutorial/add_reg_inst.rst
connor-welch: I can make changes improve the wording.

source/user/tutorial/add_proto.rst
meg-krieg: I think maybe a subsection called ``Structure and Syntax`` in the ``sim_parm.rst`` or a new  page before that could be good for this footnote . That could also be a place to mention that "hyphen" comment seen in the commodities section and stuff about capitalization mattering etc.

ConnorWelch-OSU:
Thanks for your review, @meg-krieg. I made changes to `add_reg_inst.rst` based on your comments. 

> I do still think the Add Region and Institution activity could be even less verbose. There are a lot of sequences of xml blocks to get to the general template... I am not sure how that reduction would look though.

I agree that there's still room to improve the flow here, but I think the page requires some structural changes to achieve that.

source/user/tutorial/sim_parm.rst
gonuke: Do we think the presence of these blank lines is distracting/concerning to learners?  If so, it might be best to get rid of them than to draw attention to them with this note?

source/user/tutorial/add_arche.rst
gonuke: should we get rid of these underscore characters while we're at it... the table below doesn't have them

gonuke: Curious... is there any confusion in the fact that the XML entry is `name` but we call this variable `arch1`?

Also, we end up with more archetypes in this table than in the template.  Any thoughts on rearranging this table to be something like:

+--------------+---------------+----------------+
| archetype #  |    ``lib``    |    ``name``    | 
+--------------+---------------+----------------+
|      1       |  ``cycamore`` | ``Enrichment`` |
+--------------+---------------+----------------+
|      2       |  ``cycamore`` | ``Reactor``    |
+--------------+---------------+----------------+
|      3       |  ``cycamore`` | ``Source``     |
+--------------+---------------+----------------+
|      4       |  ``cycamore`` | ``Sink``       | 
+--------------+---------------+----------------+

source/user/tutorial/add_commod_recipe.rst
gonuke: Should we remove this section or move it somewhere else?  It's an optional and advanced feature that we haven't really explained.  I think we can talk about the role of commodities without describing how or the option of adding a commodities block...

source/user/tutorial/add_reg_inst.rst
```diff
@@ -34,26 +34,26 @@ simulation.
...
+| ``arch5``   | ``NullRegion``   | Name of archetype          |
 +-------------+------------------+----------------------------+
-| ``lib7``    | ``agents``       | Library of the archetype   |
+| ``lib6``    | ``agents``       | Library of the archetype   |
 +-------------+------------------+----------------------------+
-| ``arch7``   | ``NullInst``     | Name of archetype          |
+| ``arch6``   | ``NullInst``     | Name of archetype          |
 +-------------+------------------+----------------------------+
 
 .. code-block:: XML
```
gonuke: So we need this template again, or can we just refer back to the earlier location?

source/user/tutorial/add_proto.rst
```diff
@@ -228,24 +183,6 @@ Using the template above and the table below, generate the input enrichment faci
 | ``max_feed_inventory``  | 1000000                   |
 +-------------------------+---------------------------+
 
-
-After filling in these variables, your enrichment facility prototype will look like:
-
-.. code-block:: XML
-
-  <facility>
```
gonuke, line 236: Does this mean we don't want to show the completed definition? Is that consistent across all pages?

gonuke (changes requested):
Thanks @ConnorWelch-OSU - I've got a few comments here... Look forward to your thoughts

source/user/tutorial/add_arche.rst
connor-welch: Yes, that is a good idea. I will make those changes.

connor-welch: I think that your suggestion for rearranging the table would be more clear than the current table, but I could see the potential for confusion with the archetype # column since `lib` comes before `arch` in the template. Maybe we get rid of the `lib` and `arch` numbering in the template all together and just direct the user create four `<spec>` blocks with the information in the table. What are your thoughts on that?

source/user/tutorial/add_commod_recipe.rst
connor-welch: Maybe my understanding is lacking, but don't wee need to define the commodities in order to use them as in and out commodities for prototypes?

source/user/tutorial/add_proto.rst
connor-welch: We've moved the completed definitions to the end of each page in a "Check" section.

source/user/tutorial/add_reg_inst.rst
connor-welch: We can probably exclude the template here and refer back to the archetypes page. The other option would be to instruct the user to add `NullRegion` and `NullInst` on the archetypes page.

source/user/tutorial/sim_parm.rst
connor-welch: From my personal experience going through the tutorial, having the blank lines was helpful for understanding that everything else in the tutorial was going to be put in between the control block and `</simulation>` tag

source/user/tutorial/add_arche.rst
gonuke: I noticed that in later pages, we put `[string]` in the placeholders. Maybe that's more consistent ~~and we can then replace `arch` with `name` since `name` is the XML element~~.

source/user/tutorial/add_commod_recipe.rst
gonuke: We don't need to define them.  Your understanding may actually stem from this part of the tutorial! 😀

Commodities are defined implicitly by their use in archetypes.

source/user/tutorial/add_reg_inst.rst
gonuke: I think we avoid putting these on the archetypes page because we haven't introduced the concept of regions and institutions yet.

connor-welch: That makes sense. I will remove the template here and add a reference to the archetypes page.

source/user/tutorial/add_commod_recipe.rst
connor-welch: In that case, I agree. This section isn't necessary, and may even cause confusion, as it did for me. If we were to move it, where would be the best place for it? Otherwise, I can remove it.

source/user/tutorial/add_arche.rst
connor-welch: So, for clarity, we would replace `lib1`, `arch1`, etc. in the template with `[string]`. But then, where is the `arch` that we would be replacing with `name`?

gonuke: Right! My new table suggestion had already referred to `name` so you can ignore that part of my comment

source/user/tutorial/add_commod_recipe.rst
gonuke: I think the part that introduces the concept of commodities within the DRE is useful and can stay.  Starting with the "Activity" we can remove it and add a sentence or two that states that commodities don't have to be explicitly defined. Rather, they are defined implicitly by the facilities and the commodities they choose to trade.

ConnorWelch-OSU:
Thanks for the detailed review @gonuke. I've made the following changes to address your suggestions.

- In `add_arche.rst`, I rearranged the table to remove `lib1`, `arch1`, etc. naming, but kept the columns as `Variable` and `Value` to match the way the majority of the tables in the tutorial are. Let me know what you think of the way I formatted the table.
- In `add_reg_inst.rst`, I removed the repeated template and replaced it with a link to the Understanding Archetypes page.
- I removed the optional add commodity activity, and I'm pretty sure I tracked down and removed any references to a commodity block in the rest of the tutorial, including the input files.

I also fixed the formatting for the code block and table in `add_deploy.rst` that I noticed weren't rendering properly on the website while looking for commodity references.

source/user/tutorial/add_arche.rst
```diff
@@ -138,79 +138,57 @@ where ``lib`` is the library in which the archetype came from and ``name`` is
...
-
-
-Archetype Block Template
-------------------------
++-------------+-------------+------------------+
+| Archetype # | Variable    | Value            |
++=============+=============+==================+
+| 1           | ``lib``     | ``cycamore``     |
++             +-------------+------------------+
+|             | ``arch``    | ``Source``       |
```
gonuke: There is no variable named "arch" - can we change this to "name"?

source/user/tutorial/add_commod_recipe.rst
gonuke: Should this be a template, per our normal pattern?

source/user/tutorial/add_reg_inst.rst
```diff
@@ -26,56 +26,24 @@ and institutions used in this tutorial come from the ``agents`` library, rather
...
+them into the the Archetypes block that we already have.
+Using the template on the `Understanding Archetypes <https://fuelcycle.org/user/tutorial/add_arche.html>`_ page,
+append two ``spec`` blocks into the Archetypes block for the variables listed in the table below.
+
++-------------+-------------+------------------+
+| Archetype # | Variable    | Value            |
++=============+=============+==================+
+| 5           | ``lib``     | ``agents``       |
++             +-------------+------------------+
+|             | ``arch``    | ``NullRegion``   |
```
gonuke:
```suggestion
|             | ``name``    | ``NullRegion``   |
```

gonuke (changes requested):
This is getting close, @ConnorWelch-OSU - a couple of little comments/requests here

source/user/tutorial/add_commod_recipe.rst
```diff
@@ -43,118 +43,8 @@ and provides two bids accordingly.
...
-      <solution_priority>1.0</solution_priority>
-    </commodity>
-    <commodity>
-      <name>spent_uox</name>
-      <solution_priority>1.0</solution_priority>
-    </commodity>
-
-Once complete, append the commodities section under the archetypes section [#f1]_.
+|Cyclus| does not require commodities to be defined explicitly in the input file. 
+Commodities are instead defined implicitly by facilities when they choose to trade.
```
abachma2: I understand removing this part of the tutorial. Do we have somewhere else on the website that discusses this input block and it's purpose? If so, we should add that in here as a further reading item. If we don't, we should probably add that in as documentation.

source/user/tutorial/add_commod_recipe.rst
abachma2: I agree. Let's make sure we have a template for this section somewhere on the page.

abachma2 (commented):
Thanks for continuing to work on this, @ConnorWelch-OSU. I have just a few comments, but things are looking good.

source/user/tutorial/add_commod_recipe.rst
connor-welch: As far as I can tell, the only place on the website that mentions the commodity block is in `sim_schema.rst`, but that doesn't explain it's purpose. I agree that it should exist somewhere, but I'm not sure where would be best for it. Do you have any thoughts on that?

source/user/tutorial/add_arche.rst
connor-welch: This is now fixed.

source/user/tutorial/add_reg_inst.rst
connor-welch: This is now fixed.

source/user/tutorial/add_commod_recipe.rst
connor-welch: In my most recent commits, I changed this to our template form and edited the surrounding material to fit the context of the template.

For `id` in the template, I put `[int]` as the placeholder, as that is standard format for expressing the Nuc IDs, but strings can also be used, as mentioned in lines 77 to 85. Is there a different way this placeholder should be handled given the different data types that are acceptable?

abachma2: Maybe `[string, int]`? Would that make it seem like you need to have both? Or `[string or int]`?

source/user/tutorial/add_commod_recipe.rst
abachma2: My immediate solution would be to remove the optional part from the tutorial and make a new issue for where to host the commodities input block discussion on the website. We can have a focused conversation with the new issue.

source/user/tutorial/add_commod_recipe.rst
connor-welch: I think if we want to have both in the template, something like `[string or int]` would be best. My initial thought was `[int] OR [string]`, but don't know that one is strictly better or clearer that the other.

source/user/tutorial/add_commod_recipe.rst
connor-welch: Ok, I created #446 to have that conversation.

ConnorWelch-OSU:
Thank you all for your feedback. I've added a link to https://fuelcycle.org/user/input_specs/commodity.html in the commodity section and updated the placeholder in the recipe template for `id` to `[int] OR [string]` to reflect the multiple nuclide ID formats that can be used.

source/user/tutorial/add_proto.rst
```diff
@@ -83,8 +83,7 @@ Optional parameters:
 Activity: Configure the Source prototype
 ++++++++++++++++++++++++++++++++++++++++
 Our source, ``UraniumMine``, will provide the natural uranium ore for our enrichment facility.
-This facility takes two inputs, ``name`` and ``outcommod``. Using the Source 
-Archetype and the table below, create the UraniumMine prototype.
+This facility takes two inputs, ``name`` and ``outcommod``. Using the Source archetype template above and the table below, create the UraniumMine prototype.
```
abachma2, line 86: I might be splitting hairs here, but Sink really only takes the one input, out_commod. Name is an input for facility. @gonuke can comment on if it's worth it to make that distinction here. Also, Sink can take another one or two optional parameters, and I feel like we should make it clear that this isn't the only parameter it can take.

abachma2 (commented):
This is looking really good @ConnorWelch-OSU. I have one comment that gets a bit into the finer details of the schema. @gonuke can comment on if my comment gets too detailed for the tutorial.

source/user/tutorial/add_proto.rst
gonuke: FWIW, @ConnorWelch-OSU didn't change this language, it was always there before.  I'm inclined to let it go here and consider a separate issue to correct how we introduce the concepts of "configurations" for prototypes.

source/user/tutorial/add_commod_recipe.rst
```diff
@@ -43,118 +43,9 @@ and provides two bids accordingly.
...
```
gonuke: Nitpicking on the precision of the language:
```suggestion
Commodities are instead defined implicitly by facilities when they indicate which facilities they will trade as either inputs or outputs. 
```

source/user/tutorial/add_commod_recipe.rst
```diff
@@ -164,28 +55,28 @@ Concept: Recipes
...
-     <recipe>
-      <name>nat_u</name>
-      <basis>mass</basis>
+    <recipe>
+      <name>[string]</name>
+      <basis>[string]</basis>
       <nuclide>
-        <id>92235</id>
-        <comp>0.00711</comp>
+        <id>[int] OR [string]</id>
```
gonuke, line 68: We could also just define the type as `[NucID]` and then define that later?
```suggestion
        <id>[NucID]</id>
```

source/user/tutorial/add_commod_recipe.rst
```diff
@@ -164,28 +55,28 @@ Concept: Recipes
...
-     </recipe>
+      ...
+    </recipe>
 
-where ``id`` is the Nuc Id of the isotope in form ZZAAA and ``comp`` is the
-composition of that isotope in the recipe. Other isotope formats are
+where basis can be ``mass`` or ``atom``, ``id`` is the Nuc Id of the isotope in form ZZAAA, and ``comp`` is the
+composition of that isotope in the recipe. Other isotope formats for ``id`` are
 also acceptable, such as those used by `pyne <http://pyne.io/theorymanual/nucname.html>`_. 
 For example, :math:`^{235}`\ U can be expressed as:
```
gonuke, line 81:
```suggestion
where basis can be ``mass`` or ``atom``, ``id`` identifies which nuclide in one of the forms listed below, and ``comp`` is the
composition of that nuclide in the recipe. Many formats are allowed for a ``NucID` as defined by `pyne <http://pyne.io/theorymanual/nucname.html>`_. 
For example, :math:`^{235}`\ U can be expressed as:
```

source/user/tutorial/add_commod_recipe.rst
```diff
@@ -373,5 +143,57 @@ names in cyclus and ``_`` should be used instead):
 The recipe name ``fresh_uox`` is specified, as are the isotope nuclide IDs and the 
 corresponding mass fraction of each nuclide. The ``fresh_uox`` is composed of 4% U-235 and 96% U-238.
```
gonuke: Should we delete this? It seems redundant now? (maybe always was?)

source/user/tutorial/add_proto.rst
```diff
@@ -94,34 +93,7 @@ Archetype and the table below, create the UraniumMine prototype.
 | ``out_commodity       | ``u_ore``                 |
```
gonuke: This is a long-lingering formatting error
```suggestion
| ``out_commodity``       | ``u_ore``                 |
```

source/user/tutorial/add_deploy.rst
gonuke: Should we update this template to match the others:

```suggestion
            <val>[string]</val>
```
and all other entries?

gonuke (commented):
Thanks @ConnorWelch-OSU - I'm trying not to make this take for ever, but found a few more little issues that were not entirely consistent.

I have viewed the rendered version that the CI testing makes, and in general it all looks great.

source/user/tutorial/add_commod_recipe.rst
connor-welch: That is an option, but we would need to decide if we're ok with not using the data type as the placeholder in this one instance.

source/user/tutorial/add_commod_recipe.rst
connor-welch: I tend to agree. I will remove it.

source/user/tutorial/add_commod_recipe.rst
gonuke: Arguably - it is the data type...

connor-welch: I guess by data type, I meant string, int, double, bool, etc.

gonuke (approved):
Thanks for this thorough review @ConnorWelch-OSU - I think we should merge this now and we can continue to fine tune in subsequent issues and PRs, lest this sit here forever with every new idea we have.

PR #447, opened 2026-02-20
CEP 31 - Agent Registered Discrete Event Timing  

meg-krieg:
# Summary of Changes

This PR adds a CEP 31 for discrete event timing proposal in Cyclus (and by extent cycamore). There are many areas where a more robust implementation can be proposed including 

1) treatment of Tick/Tock/Decision functions
2)  required ```cycle_length``` for each archetype to schedule new events under a capacity-met condition
3) maintaining the same ```Timer::RunSim()``` phase suite ordering within each event even if timesteps are skipped

The CEP goes through a broad restructuring ideas and more detailed implementation could be outlined in the future. Any suggestions to the implementations, clarifications to the CEP, or other comments from the cyclus community would be great! 

# Related CEPs and Issues
Creates new CEP 31.

### DIFF
--- a/source/cep/cep31.rst
+++ b/source/cep/cep31.rst
@@ -43,12 +43,13 @@ within the simulation duration. The simulation progresses from scheduled event t
 where no actions are registered. For example, an agent may schedule a build-event at timestamp 5, a trade event at timestamp 7, and a decomission event at timestamp 10.
 Then, the timeline for a 10 month simulation will proceed as follows 
 
-
-START 3 (build) -- 5 (trade) -- 7 (decom) END
+centered:: 
+  START 3 (build) -- 5 (trade) -- 7 (decom) END
 
 as opposed to the current implementation 
 
-START 0 -- 1 -- 2 -- 3 (build) -- 4 -- 5 (trade) -- 6 -- 7 (decom) -- 8 -- 9 -- 10 END
+centered:: 
+  START 0 -- 1 -- 2 -- 3 (build) -- 4 -- 5 (trade) -- 6 -- 7 (decom) -- 8 -- 9 -- 10 END
 
 In |cyclus|, discrete even timing can be implemented by allowing agents to internally check their inventory
 and status to register themselves for DRE participation, Build, or Decomission events. In instances
@@ -63,7 +64,7 @@ Conceptualization
 Based on the current structure set up by the kernal and agent steps in the phase suit, the following
 discrete event implementation is suggested...
 
-Events in the simulation's internal clock can be requested by agents using their own `EventRequest()` function. 
+Events in the simulation's internal clock can be requested by agents using their own ``EventRequest()`` function. 
 Events may only be registered when agents wish to complete the follwing actions: 
 
 - Build
@@ -91,16 +92,16 @@ for all available agents as well. These internal Tick/Tock/Decision updates are 
 filtering options that consider deregistering an agent's Tocks/Decisions when they make no material bids. 
 
 Implementation
-+++++++++++++++++++++++
+===============================
 
 Material requests, Build, and Decomissioin event registration will be handled by individual agents. |Cyclus| already creates a preconditioned timeline 
-for discrete-build and decomission events. An additional `EventRequest()` member funcition for all |cycamore| archetypes will check a facility's inventory.  
+for discrete-build and decomission events. An additional ``EventRequest()`` member funcition for all |cycamore| archetypes will check a facility's inventory.  
 
 1. If inventory not at capacity, agent will register for the (+1) next immediate time step to attempt another request.
-2. If inventory at capacity, agent will register its next request event for a fixed `+ cycle_length` time in from the current event. 
+2. If inventory at capacity, agent will register its next request event for a fixed ``+ cycle_length`` time from the current event. 
 
 These material request events will be registered within Context in a dynamic dictionary that contains the event's timestamp and a list of ``Trader`` objects. To ensure that all
-facility agents' `EventRequest()` functions are checked regularly, A look-ahead function will be added to |cyclus's| cardinal phase suite after the decomissioning phase. 
+facility agents' ``EventRequest()`` functions are checked regularly, A look-ahead function will be added to |cyclus's| cardinal phase suite after the decomissioning phase. 
 
 .. code-block:: c++
     DoBuild();
@@ -112,9 +113,9 @@ facility agents' `EventRequest()` functions are checked regularly, A look-ahead 
     DoLookAhead();
 
 The cardinal phase suite will maintain its current ordering for the reasons described in ---. Each of the 6 phases will be checked during each event (as opposed to each time step) but the phase will 
-only be triggered and completed if it has participants registered. Only the newly introduced `DoLookAhead()` that checks each agents `EventRequest()` will be executed for all agents each event. 
+only be triggered and completed if it has participants registered. Only the newly introduced ``DoLookAhead()`` that checks each agents ``EventRequest()`` will be executed for all agents each event. 
 
-The simulation will be terminated fully when `DoLookAhead()` registers no new events and the simulation has completed the last event registered.
+The simulation will be terminated fully when ``DoLookAhead()`` registers no new events and the simulation has completed the last event registered.
 
 Backwards Compatibility
 ========================
--- a/source/cep/cep31.rst
+++ b/source/cep/cep31.rst
@@ -1,0 +1,127 @@
+CEP 31 - Agent Registered Discrete Event Timing 
+***********************************************************
+
+:CEP: 31
+:Title: Agent Registered Discrete Event Timing 
+:Last-Modified: 2026-02-19
+:Author: Meghan Krieg <kriegm@oregonstate.edu>
+:Status: Draft
+:Type: Standards Track
+:Created: Paul Wilson
+
+Background
+===========
+
+|Cyclus| currently runs using a fixed increment time advancement mechanism, ``dt``. 
+Within each ``dt`` of the simulation, |cyclus's| cardinal function progresses through a distinct ordering
+of 6 phases which include the 
+
+- Building Phase
+- Tick Phase 
+- Exchange Phase 
+- Tock Phase
+- Decision Phase
+- Decomissioning Phase
+
+before advancing to the next time step. One of the core paradigms of this implementation is 
+that all agents within the simulation must participate in the Dynamic Resource Exchange and 
+the surrounding time execution steps (Tick/Tock phases). The ordering of these phases allows
+the maxiumum transition of resources with building phase completed first and decomissioning last. 
+In addition to this feature, the phase ordering toggles between two categories: agent phase nad kernel phase. 
+While the kernel phase represents events that alter the simulation state, agent phases include
+actions that update an agent's internal state. For more information, refer to ---. 
+
+Where --- defined |cyclus| as a broadly discrete event simulation, CEP 31 will push the implementation
+further. 
+
+Motivation and Rationale
+==========================
+
+Simulations with event based or discrete event timing run on an internal clock rather than a fixed simulation
+clock (as featured in |cyclus|). For this internal clock, events that alter the simulation state are registered at distinct timestamps
+within the simulation duration. The simulation progresses from scheduled event to scheduled event skipping timestamps
+where no actions are registered. For example, an agent may schedule a build-event at timestamp 5, a trade event at timestamp 7, and a decomission event at timestamp 10.
+Then, the timeline for a 10 month simulation will proceed as follows 
+
+
+START 3 (build) -- 5 (trade) -- 7 (decom) END
+
+as opposed to the current implementation 
+
+START 0 -- 1 -- 2 -- 3 (build) -- 4 -- 5 (trade) -- 6 -- 7 (decom) -- 8 -- 9 -- 10 END
+
+In |cyclus|, discrete even timing can be implemented by allowing agents to internally check their inventory
+and status to register themselves for DRE participation, Build, or Decomission events. In instances
+when no agents reigster actions, there will be no events, and the simulation will skip that timestmap. Agents will
+dicate the dynamics of the simulation instead of a fixed timestep forcing interactions. 
+
+With many timesteps skipped between agent's cycle/event registrations, computational time will be lower. 
+
+Conceptualization 
+===============================
+
+Based on the current structure set up by the kernal and agent steps in the phase suit, the following
+discrete event implementation is suggested...
+
+Events in the simulation's internal clock can be requested by agents using their own `EventRequest()` function. 
+Events may only be registered when agents wish to complete the follwing actions: 
+
+- Build
+- Request Materials
+- Decomission 
+
+These actions map to the current *kernel* phases of |cyclus's| cardinal function. Thus, by extent, 
+
+- Tick
+- Tock
+- Decision
+
+may not be considered events because they do *not* change the simulation state. They change 
+the agent's internal state (agent phase). As these actions are not events, they may never be independently registered by agents. 
+(There are scenarios in which build/requets/decomission may trigger them, however.)
+
+For the DRE, agents will *only* be able to register for material requests; thus, material demands instead of material
+supply will drive the DRE scheduling. When agents go to register an event by requesting a material, they are unaware of the state of other agent's 
+supplies. So, for the subset of agents that are triggering DRE events through material requests, *all available
+agents in the simulation must be accessible to the DRE to request bids and responses*. This ensures that material request events aren't prohibited from being satisfied 
+even when requested material exists. 
+
+Because of the required attendance for all agents in the material bids stage of the DRE, the agent phases Tick/Tock/Decision must be triggered 
+for all available agents as well. These internal Tick/Tock/Decision updates are crucial for deriving accurate *bid* information for request portfolios. There are further
+filtering options that consider deregistering an agent's Tocks/Decisions when they make no material bids. 
+
+Implementation
++++++++++++++++++++++++
+
+Material requests, Build, and Decomissioin event registration will be handled by individual agents. |Cyclus| already creates a preconditioned timeline 
+for discrete-build and decomission events. An additional `EventRequest()` member funcition for all |cycamore| archetypes will check a facility's inventory.  
+
+1. If inventory not at capacity, agent will register for the (+1) next immediate time step to attempt another request.
+2. If inventory at capacity, agent will register its next request event for a fixed `+ cycle_length` time in from the current event. 
+
+These material request events will be registered within Context in a dynamic dictionary that contains the event's timestamp and a list of ``Trader`` objects. To ensure that all
+facility agents' `EventRequest()` functions are checked regularly, A look-ahead function will be added to |cyclus's| cardinal phase suite after the decomissioning phase. 
+
+.. code-block:: c++
+    DoBuild();
+    DoTick();
+    DoResEx(&matl_manager, &genrsrc_manager);
+    DoTock();
+    DoDecision();
+    DoDecom();
+    DoLookAhead();
+
+The cardinal phase suite will maintain its current ordering for the reasons described in ---. Each of the 6 phases will be checked during each event (as opposed to each time step) but the phase will 
+only be triggered and completed if it has participants registered. Only the newly introduced `DoLookAhead()` that checks each agents `EventRequest()` will be executed for all agents each event. 
+
+The simulation will be terminated fully when `DoLookAhead()` registers no new events and the simulation has completed the last event registered.
+
+Backwards Compatibility
+========================
+
+These changes will not be backwards compatible with |cyclus| v1.6.0 and may require a new release. An optional simulation parameter could be introduced to switch this treatment on and off 
+such that the original time implemenation can be used.
+
+
+
+
--- a/source/cep/cep31.rst
+++ b/source/cep/cep31.rst
@@ -5,15 +5,17 @@ CEP 31 - Agent Registered Discrete Event Timing
 :Title: Agent Registered Discrete Event Timing 
 :Last-Modified: 2026-02-19
 :Author: Meghan Krieg <kriegm@oregonstate.edu>
+:BDFP: Madicken Munk/Paul Wilson
 :Status: Draft
 :Type: Standards Track
-:Created: Paul Wilson
+:Created: 2026-02-20
+:Cyclus-Version: 1.6.0
 
 Background
 ===========
 
 |Cyclus| currently runs using a fixed increment time advancement mechanism, ``dt``. 
-Within each ``dt`` of the simulation, |cyclus's| cardinal function progresses through a distinct ordering
+Within each ``dt`` of the simulation, cyclus's cardinal function progresses through a distinct ordering
 of 6 phases which include the 
 
 - Building Phase
@@ -29,29 +31,29 @@ the surrounding time execution steps (Tick/Tock phases). The ordering of these p
 the maxiumum transition of resources with building phase completed first and decomissioning last. 
 In addition to this feature, the phase ordering toggles between two categories: agent phase nad kernel phase. 
 While the kernel phase represents events that alter the simulation state, agent phases include
-actions that update an agent's internal state. For more information, refer to ---. 
+actions that update an agent's internal state. For more information, refer to `CEP 20<https://fuelcycle.org/cep/cep20.html>`_. 
 
-Where --- defined |cyclus| as a broadly discrete event simulation, CEP 31 will push the implementation
+Where `CEP 20<https://fuelcycle.org/cep/cep20.html>`_ defined cyclus as a broadly discrete event simulation, CEP 31 will push the implementation
 further. 
 
 Motivation and Rationale
 ==========================
 
 Simulations with event based or discrete event timing run on an internal clock rather than a fixed simulation
-clock (as featured in |cyclus|). For this internal clock, events that alter the simulation state are registered at distinct timestamps
+clock (as featured in cyclus). For this internal clock, events that alter the simulation state are registered at distinct timestamps
 within the simulation duration. The simulation progresses from scheduled event to scheduled event skipping timestamps
 where no actions are registered. For example, an agent may schedule a build-event at timestamp 5, a trade event at timestamp 7, and a decomission event at timestamp 10.
 Then, the timeline for a 10 month simulation will proceed as follows 
 
 .. class:: center
-  START 3 (build) -- 5 (trade) -- 7 (decom) END
+START 3 (build) -- 5 (trade) -- 7 (decom) END
 
 as opposed to the current implementation 
 
 .. class:: center
-  START 0 -- 1 -- 2 -- 3 (build) -- 4 -- 5 (trade) -- 6 -- 7 (decom) -- 8 -- 9 -- 10 END
+START 0 -- 1 -- 2 -- 3 (build) -- 4 -- 5 (trade) -- 6 -- 7 (decom) -- 8 -- 9 -- 10 END
 
-In |cyclus|, discrete even timing can be implemented by allowing agents to internally check their inventory
+In cyclus, discrete even timing can be implemented by allowing agents to internally check their inventory
 and status to register themselves for DRE participation, Build, or Decomission events. In instances
 when no agents reigster actions, there will be no events, and the simulation will skip that timestmap. Agents will
 dicate the dynamics of the simulation instead of a fixed timestep forcing interactions. 
@@ -71,7 +73,7 @@ Events may only be registered when agents wish to complete the follwing actions:
 - Request Materials
 - Decomission 
 
-These actions map to the current *kernel* phases of |cyclus's| cardinal function. Thus, by extent, 
+These actions map to the current *kernel* phases of cyclus's cardinal function. Thus, by extent, 
 
 - Tick
 - Tock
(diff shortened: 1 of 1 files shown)


source/cep/cep31.rst
```diff
@@ -0,0 +1,128 @@
...
+
+These actions map to the current *kernel* phases of cyclus's cardinal function. Thus, by extent, 
+
+- Tick
+- Tock
+- Decision
+
+may not be considered events because they do *not* change the simulation state. They change 
+the agent's internal state (agent phase). As these actions are not events, they may never be independently registered by agents. 
+(There are scenarios in which build/requets/decomission may trigger them, however.)
```
gonuke, line 82: This seems like a key point of discussion/contention.  Can you say more about why you want this strict interpretation? If the context will be skipping time steps based on the registered events, how will agents know when it is time to tick or tock? How will they be triggered? What if a tick is necessary to determine whether or not to participate in a trade?

source/cep/cep31.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Because of the required attendance for all agents in the material bids stage of the DRE, the agent phases Tick/Tock/Decision must be triggered 
+for all available agents as well. These internal Tick/Tock/Decision updates are crucial for deriving accurate *bid* information for request portfolios. There are further
+filtering options that consider deregistering an agent's Tocks/Decisions when they make no material bids. 
+
+Implementation
+===============================
+
+Material requests, Build, and Decomissioin event registration will be handled by individual agents. Cyclus already creates a preconditioned timeline 
+for discrete-build and decomission events. An additional ``EventRequest()`` member funcition for all cycamore archetypes will check a facility's inventory.
```
gonuke, line 98: If such a function is necessary, it should perhaps be added to the `agent` class from which these all derive.

source/cep/cep31.rst
```diff
@@ -0,0 +1,128 @@
...
+filtering options that consider deregistering an agent's Tocks/Decisions when they make no material bids. 
+
+Implementation
+===============================
+
+Material requests, Build, and Decomissioin event registration will be handled by individual agents. Cyclus already creates a preconditioned timeline 
+for discrete-build and decomission events. An additional ``EventRequest()`` member funcition for all cycamore archetypes will check a facility's inventory.  
+
+1. If inventory not at capacity, agent will register for the (+1) next immediate time step to attempt another request.
+2. If inventory at capacity, agent will register its next request event for a fixed ``+ cycle_length`` time from the current event.
```
gonuke, line 101: This is too rooted in the current set of archetypes.  We need to think beyond the concept of cycle length and reactors.

source/cep/cep31.rst
```diff
@@ -0,0 +1,128 @@
...
+===============================
+
+Material requests, Build, and Decomissioin event registration will be handled by individual agents. Cyclus already creates a preconditioned timeline 
+for discrete-build and decomission events. An additional ``EventRequest()`` member funcition for all cycamore archetypes will check a facility's inventory.  
+
+1. If inventory not at capacity, agent will register for the (+1) next immediate time step to attempt another request.
+2. If inventory at capacity, agent will register its next request event for a fixed ``+ cycle_length`` time from the current event. 
+
+These material request events will be registered within Context in a dynamic dictionary that contains the event's timestamp and a list of ``Trader`` objects. To ensure that all
+facility agents' ``EventRequest()`` functions are checked regularly, A look-ahead function will be added to cyclus's cardinal phase suite after the decomissioning phase.
```
gonuke, line 104: Doesn't this degenerate to time steps again?  I don't know what the purpose of a look-ahead function is and when/how it gets triggered?

gonuke (changes requested):
Thanks for kicking this off @meg-krieg!

I'm not sure I understand the motivation for not allowing Tick/Tock/Decision to be registered as events that come at discrete time queued by the context/timer?

source/cep/cep31.rst
meg-krieg: I pasted some general code below in case my words are really vague! Maybe this makes what I am thinking a little clearer? It is pretty bare-bones but just for a general idea -- in ``timer.cc``

 ```c++
void Timer::RunSim() {
...
ctx_->Populate(0); // at time 0 everyone must register for a trade event to get started
  while ( (time_ < si_.duration) && (prev_time_ != time_)) {

    CLOG(LEV_INFO1) << "Current time: " << time_;
    if (want_snapshot_) {
      want_snapshot_ = false;
      SimInit::Snapshot(ctx_);
    }
    // run through phases
    DoBuild();
    CLOG(LEV_INFO2) << "Beginning Tick for time: " << time_;
    DoTick();
    CLOG(LEV_INFO2) << "Beginning DRE for time: " << time_;
    DoResEx(&matl_manager, &genrsrc_manager);
    CLOG(LEV_INFO2) << "Beginning Tock for time: " << time_;
    DoTock();
    CLOG(LEV_INFO2) << "Beginning Decision for time: " << time_;
    DoDecision();
    DoDecom();
    DoLookAhead();

#ifdef CYCLUS_WITH_PYTHON
    EventLoop();
#endif
    prev_time_ = time_;
    time_ = NextEvent();
...
}
```

Then, later in ``timer.cc``, something similar to this will happen where timer checks for the next closest event registered in the timeline. (this is not meant to be accurate code). 
```c++
void Timer::DoLookAhead() {
  std::set<Trader*> all_traders = ctx_->traders();
  for(Trader* m : all_traders){
    m->EventRequest(); 
  };
// another check will probably go here
}

int Timer::NextEvent(){
  auto reg_traders = ctx_->EventRequesters(); //the list of traders who have registered for events in the timeline
  int t_p = time_ +1; // time plus +1 
  std::vector<int> event_lists = {decom_queue_.upper_bound(t_p)->first,build_queue_.upper_bound(t_p)->first}, reg_traders.upper_bound(t_p)->first};
  return *std::min_element(event_lists.begin(), event_lists.end());
}
```
So, the simulation finishes some event and it looks for the next closest timestamp that has an event. That timestamp could have events registered under any 3, 2, or 1 of build/trade/decom actions scheduled. The simulation knows it has the event at the timestamp but does not know which type (this could change maybe), so it checks each phase for that event.  If the phase is filled with participants, that phase is one of the events and is triggered. 

This means that the phases will have some conditional check... for example 

```c++
void Timer::DoBuild() {
  if(build_queue_.count(time_)==0){
         continue;}
else{
  std::vector<std::pair<std::string, Agent*>> build_list = build_queue_[time_];
  for (int i = 0; i < build_list.size(); ++i) {
    Agent* m = ctx_->CreateAgent<Agent>(build_list[i].first);
    Agent* parent = build_list[i].second;
...
   }
...
  }
```
Overall, this treatment should skip the timeline from event to event instead of timestep to next timestep when fully fleshed out.

source/cep/cep31.rst
meg-krieg: There are scenarios in which i can see the Tock and Decision being registrable events based on the outcome of the DRE. 

 It is mainly the tick that cannot be registrable (in my opinion)! If all traders objects are available bidders every trade event and all bidders need accurate bid information and we don't know who will bid, then all bidders must all tick. In other words, all traders must tick either every scheduled event or every trade event. 

I could be really off the mark though.

meg-krieg: I have come up with a better way to articulate this idea. I think Tock/DoDecision are secondary events whereas Build/DoResEx/Decom are primary events. Secondary events can only occur when a primary event has been completed. This means that the occurrence of a trading event informs whether a Decision/Tock should occur at all. 

So, I was thinking that agents only register primary events, and primary events may or may not schedule secondary events. Ideally, I was hoping to relocate any building/decom/trading actions that have leaked into Tock/Decision to make this possible. For example, I know it is the Institution::Tock that schedules agent decommissioning...I would hope to move this action and related decom/build/material transfer actions from the agent phases and keep agent phases more about an agent's internal inventory/updates.

So in this sense, Tock/Decision would be scheduled within an current timestamp (not in the future) for agents who have completed a trade. Ticks I still think should occur for every trader prior to a registered trade event. But, is there a specific instance of a tick where you could see this being a big problem?

source/cep/cep31.rst
```diff
@@ -0,0 +1,128 @@
...
+- Exchange Phase 
+- Tock Phase
+- Decision Phase
+- Decomissioning Phase
+
+before advancing to the next time step. One of the core paradigms of this implementation is 
+that all agents within the simulation must participate in the Dynamic Resource Exchange and 
+the surrounding time execution steps (Tick/Tock phases). The ordering of these phases allows
+the maxiumum transition of resources with building phase completed first and decomissioning last. 
+In addition to this feature, the phase ordering toggles between two categories: agent phase nad kernel phase.
```
abachma2, line 32:
```suggestion
In addition to this feature, the phase ordering toggles between two categories: agent phase and kernel phase. 
```
small typo

abachma2 (commented):
Thanks for putting this together, @meg-krieg. I think you have some solid ideas, and the presentation you gave yesterday at the Cyclus call was very helpful. 

I've been thinking about the conversation yesterday regarding the utility of having bids trigger events sometimes, and how only having requests trigger events may lead to trying to fill large buffers that don't need to be filled. Is there a way to allow requests and bids trigger an event, like a `TradeEvent` function to register when one or both events need to happen, then at the archetype level we can define which trade portfolio gets used in that function to decide when to trigger an event? That we can could have front-end facilities allow their requests to trigger events, reactors can have both trigger events, and the Storage or Sink archetypes can have neither trigger events. I think this gives some flexibility to archetype developers.

source/cep/cep31.rst
meg-krieg: Thanks for the comment -- I like the idea of having something to work around the storage and sink triggering whenever they have space for materials!!

Would the presence of any material to bid/supply trigger an DRE event? Or would it be a threshold of material exceeded that would trigger the DRE event? If I am understanding correctly. you mention only the reactor could have the bidding event registration ability. Should the enrichment and separations facility who also produce some sort of material for storage/sink also have this capability? 

Then, similar to how all bidders must be present for any subset of facilities requesting, any requester must be present for any subset of bidders bidding to ensure that the bids are met.

abachma2: I think there are various ways you could do this. You could have any bid, request, or either for a facility trigger an event, you could have it only trigger if that ResBuf is getting full (maybe when the remaining capacity is less than the previous material intake?). 

The main idea is to be flexible, allowing the archetype developer define what type of material transaction triggers an event. There may be cases in which a developer wants only bids to trigger events. Your CEP would provide flexibility, and then we would need to decide as a community which DRE portfolio should trigger events for the Cycamore archetypes. I just listed those archetypes as examples. 

You would need to decide if a full DRE takes place when triggered by the `TradeEvent` (or whatever you call it) and any facility that has a bid/request portfolio participates, or only those with portfolios involving the commodity(ies) that triggered the event -- I would recommend the former. 

(a bit stream of conscious here, I apologize) Also, would a full time step take place when any event is triggered? Would the DRE take place when a build or decommission triggers an event?

meg-krieg: No, the DRE would not be triggered when a build or decommission triggers.  pre-DRE, there will be a check for DRE participants and if there are none it is not triggered. In response to the other part of the question, a full timestep is occurring within the event as we had some discussion about Ticks/Tocks needing to sandwich any event type (DRE/build/decom). I am going to start a discussion in a different thread about that! 

And thanks for the clarification, I see that having some sort of future developer be able to decide which events (request/bid/both etc) trigger DRE is important. I think one of the questions this prompts is the future time for which the event is scheduled. With this flexibility. it would not only be on the developer to select the event to be registered but also select when the event occurs. For example, a developer could decide they want their archetype to only trigger DRE with bids but is that developer also deciding some function/model the predicts the time to schedule the bids? I hope if what I am asking makes sense. 

And I think i agree that the entire DRE should happen given the triggered trade event, but I think that Paul was speculating it is only possible when all requesters for a subset of bids or all bidders for a subset of requests must be present in order for the DRE to be "reliable."

meg-krieg:
## Tick and Tock discussion continued

From the conversation last Tuesday, I wanted to recap some of the ideas discussed at the end about the broad implementation of discrete event in cyclus (beyond cycamore). I have written a short blurb about what I remember being said but if I misinterpreted or remembered incorrectly, please correct me! 

in my pressentation, I noted that Ticks/Tocks were a point of contention. I suggested three scenarios in which 
1) Tick/Tock happen every event and is not scheduled by agents but automatically applied to all agents.
I suspect that this would be the least invasive to cyclus but also more straightforward implementation ^. But, I am biased. 

2) Ticks/Tocks are treated as secondary events so they are scheduled by agents *within* a timestamp as a result of primary event (DRE/decom/build) scheduling
3) Hybrid: Ticks occur for every agent for every event, but Tocks an be registered like secondary events.  Example being an agent enters DRE but completes no trade, so a Tock is unneeded. 

 @gonuke  thought that agents should be able to schedule any event type (Tick/Tock/MaterialReqs/Build/Decom) whenever they need to. Tick/Tock are not secondary events or tied to any other process. Allowing this means that the agent has more control about when it needs to know about the simulation state. 

@katyhuff mentioned that Ticks and Tocks should occur together for any/every event as they sandwich kernel phases in cyclus. It was also suggested that any ``Do`LookAhead()``` function (that is called to allow all agents to assess whether they need to register themselves in the DRE) should be moved to the Tock. As a result, if the scheduling for DRE is occurring in Tocks it would support the need for Tocks to happen every event. 

At the end of the discussion, we thought that maybe investigating further whether there is any scenario where Ticks and Tocks could be registered independent for each other and if any decom/DRE/build event only warrants one or the other. 

Does this seem like an accurate summary and also next steps that concluded the call?

gonuke:
Thanks for this summary @meg-krieg.

As a follow on, I've been thinking more about the definition of "kernel phases" and "agent phases" offered in CEP20.

First, it appears that none of our archetypes (cycamore) have implemented the `Decision` method, and I wonder if:
1. they don't need to make decisions in the way we expected, or
2. they make such decisions at other times (e.g. Tick or Tock?)

Second, it makes sense to me to offer a different categorization:
* actions that depend **only** on the internal state of the agent, vs
* actions that depend on an agent's interaction with the ecosystem

This seems useful to me because I presume that an agent can predict its internal state well and therefore can know when it needs to perform any action depending only on that state in the future, and therefore, can reliably add such a future event to the event queue. It may also be argued that such events never need to be added to the queue because we can allow the agent to leap beyond any future times when it doesn't need to interact because we can predict the state in any of those future times.

Conversely, an agent has no ability to predict how the ecosystem will change, so needs to be more careful about when to schedule such events.  It can predict when it may need to interact with the ecosystem (e.g. acquire new material), but it can't predict the outcome of those interactions (will material be available).  Therefore, an agent will need to use some strategy to determine when to schedule such events.

I am not sure how exactly this framing impacts the design, but wanted to note it here.

What I heard from @katyhuff's, and found somewhat compelling, is that the primary purpose of a Tick is to prepare internally for trades and the primary purpose of a Tock is to follow-up internally after a trade.  If we also view Ticks/Tocks as depending only on the internal state of the agent, then there is an argument that they never need to occur except for when it's time to prepare for a trade.  That is, there is no need to update the internal state of the agent except to prepare for a trade if we can always predict the future state of the agent.  This may be too limiting, but perhaps a useful ideal framing.

So one approach, and one that may also be a good starting point regardless of where things go, is that agents only ever insert future trades into the event queue, and then perform Tick/Trade/Tock at each moment of time when they request a trade.  It seems that during a Tock is the right time for an agent to therefore insert itself into the queue.

It does still require that all possible producers perform the same actions, which may again clutter the timeline with activity we are aiming to reduce, but that may be life.

meg-krieg:
I think I agree for the starting point of having Tick/Trade/Tock for every registered trade event is a must. But I can also imagine scenarios for the other 2-3 events, Build and Decom (and my allow agent to respond to simulation state function) that may also require the Tick/Tock sandwich. 

For example, 
1) *Build*: Agent is built and it should probably enter itself immediately into a trade to start getting material for its empty inventories. Thus, agent will register for trade and by extent Tick/Tock.

2) *Decom*: Agent is decommissioning and it needs to know if it has any inventory to get rid of etc. It will need to Tick to assess this and Tock as part of the sandwich. We *maybe* could get around this. 

3) Agent responds to simulation state in Tock or another function: 

If after events Trade/Build/Decom,  agents need to know when they are doing something next (if they haven't already scheduled something in the far future instead of immediate), they will need to assess their internal state to schedule any events based on what has happened previously in the simulation. If this function is in the Tock then I think there automatically needs to be a Tick. 

So in other words, 

> It seems that during a Tock is the right time for an agent to therefore insert itself into the queue.

In previous discussions of this CEP, this was my ''DoLookAhead()' function; however, having some process for inserting into the queue executed through the Tock is essentially the same thing. I wanted to have the agent assess their internal state in a Tock after any event because this means that all agents are responding to any change in the cyclus ecosystem. Thus, any ecosystem change could prompt a new scheduled event from anyone present. This will be clunky but will always make sure the agent is responding to the ecosystem changes through Tock (and by extent a Tick is needed). 

But, I can see your point that this may be overkill and perhaps it is some evolution of this idea with a little more nuance but the same structure? 

In a different thread with Amanda,  we are discussing whether the scheduling function unique to each archetype is on the developers to decide. This means that the developers would have flexibility for how their facility wants to schedule DRE related events either through bids/requests/both and possibly the time scheduling function (developers also decide when the event happens).

gonuke:
It seems worth mentioning/recalling that with this change there would no longer be any time steps.  The only opportunity for anything to happen will be at instances in time that agents have registered some activity to happen. 

There will be a time resolution (seconds?hours?days?) but there will be no opportunity to regularly invoke any method that will do something like look ahead.  

So each agent can decide to insert its next event into the queue at any time. We don't need to decide when this happens.  We can suggest guidance, best practice, patterns, but we don't need to prescribe anything about when this happens.

It makes sense to me that when a facility is built, it will insert a first event into the queue to try to acquire material. Even if we don't formally require it, it is probably a de facto requirement because based on my above discussion, there will be no other opportunity to insert an event into the queue.

gonuke:
> Thus, any ecosystem change could prompt a new scheduled event from anyone present. This will be clunky but will always make sure the agent is responding to the ecosystem changes through Tock (and by extent a Tick is needed). 

In my paradigm, I suggested that a Tock only respond to the internal state of the agent and its own predictions about its future state.  Since it occurs after a trade, the agent's state will be updated based on the state of the ecosystem, but there is no way for a Tock to learn anything the ecosystem.

Thus, if an agent thinks it needs to probe the state of the ecosystem at some time in the future, it needs to predict when it needs to do this and schedule an event to happen then.

meg-krieg:
I totally agree with you that 

> there would no longer by any timesteps .. only instances in time that agents have registered. 

 My implementation idea should align closely with this as it does not depend on timesteps but *timestamps* which I believe you are labeling as instances. Instances/timesteps are dynamically scheduled (except for facility building as those are predetermined).

Then, I also agree with you that 
>there is no way for a Tock to learn anything about the ecosystem. 

I was trying to convey how in some pre-encoded sequence of actions, a Tock should be triggered to always follow some (scheduled instance of) build/trade event, not that a Tock itself is learning about the ecosystem. I poorly worded that. 

Finally, for the ``DoLookAhead``, I agree it is not regularly occurring in the sense that it is *not* happening at a set interval timestep or cadence. For example, it is not scheduled every 6 months etc.  Instead, I was suggesting that it occurs whenever an irregular/dynamically scheduled event has been completed. I sort of think of it like dynamically scheduling some *event-scheduling method* to occur with the completion of an event... But, maybe you still would call this regular??

For clarify, ``DoLookAhead`` was meant to invoke an event scheduling method that agents have. ``DoLookAhead`` does not schedule events itself.  

Perhaps, these two separate sets of implementation ideas are more converged than expected?? .. But, I may be using the wrong language to explain my thoughts and misinterpreting.

gonuke:
I think we're converging on some ideas here.  Here's what I am seeing emerge:

1. any agent can add itself to the execution queue for any timestamp in the future
    * I think we're converging towards this happening during a Tock (or a Build when starting up), but we can come back to that
3. whenever an agent acts, it must Tick/Trade/Tock
    * this is presumably consumer-driven; there is no reason for a supplier to trigger action
4. all suppliers must be available any time that a consumer is acting
    * they must also perform a Tick/Trade/Tock
5. all consumers of the same commodity should always act at the same time

I think the consequence of this is that we maintain a queue of commodity trade events. That is, when an agent wants to register an event, it's actually registering that a given commodity needs to be traded and all agents connected to that commodity as either consumer or supplier need to act during that time step with a Tick/Trade/Tock cycle.

meg-krieg:
So far yes, I think that 1-3 is where the discussion has landed, and I agree in instances it could be similar to original Cyclus. 

When you say that consumers of the commodity should always act at the same time in bullet 4, why would this be needed? Is this just another feature of that described implementation or .... 

Is this intended to be a transition from 

"consumer registers that they need a commodity and all suppliers (regardless of commodity output) enter to meet that request"  (what we have discussed so far) 

to 

"trade events are registered either through request or bidding (of any commodity) and only facilities related to that commodity (suppliers or consumers) enter the DRE. The request or bid of a commodity triggers an event" Amanda had been discussing something similar for bids being necessary for events especially for back end FC.

meg-krieg:
Because there are concerns about the above implementation being very similar to original cyclus, I am just going to speculate outside of the box and introduce an entirely different scheme. Since the reactor is the only facility with a cycle length, the "scarcity" in the spent fuel supply and rarity of the fresh fuel demand seem to dictate a lot of the simulation dynamics. This means that beyond build/decomission events, most facilities stay idle until the reactor state is changing. 

There could be an implementation where only reactors trigger simulation wide trade events were the event (single) starts when the reactor ends cycle or discharges fuel and event ends when reactor starts a new cycle (has new fuel). 
Our event set would change from build/decom/request to build/decom/reactor refueling. Then, for the other facilities we could solve for some end state (difficult for complex FC) of all the facilities when the reactor is satisfied and entering a new cycle. For simple FC example, if a reactor discharges 3 assemblies of X mass, we know each facility will be moved to produce or accept some cumulative amount over some whole refueling period event (based on possible throughputs and capacities of course). The end-refueling state of a facilities will be stored and practically unchanging until  at the next event (build/decom/refueling) that stored state will be "uploaded."

In effect this would be cycle driven discrete dynamics and since reactors are the only ones with cycles -- reactor driven discrete dynamics. The timeline of events will be preconfigured almost entirely. 

*This is most likely not worth while idea since it removes a lot of Cyclus's flexibility.*

If you squint, this structure is already a dynamic of the current CEP's ideas.

gonuke:
> Because there are concerns about the above implementation being very similar to original cyclus, I am just going to speculate outside of the box and introduce an entirely different scheme. Since the reactor is the only facility with a cycle length, the "scarcity" in the spent fuel supply and rarity of the fresh fuel demand seem to dictate a lot of the simulation dynamics. This means that beyond build/decomission events, most facilities stay idle until the reactor state is changing.
> 
> There could be an implementation where only reactors trigger simulation wide trade events were the event (single) starts when the reactor ends cycle or discharges fuel and event ends when reactor starts a new cycle (has new fuel). Our event set would change from build/decom/request to build/decom/reactor refueling. Then, for the other facilities we could solve for some end state (difficult for complex FC) of all the facilities when the reactor is satisfied and entering a new cycle. For simple FC example, if a reactor discharges 3 assemblies of X mass, we know each facility will be moved to produce or accept some cumulative amount over some whole refueling period event (based on possible throughputs and capacities of course). The end-refueling state of a facilities will be stored and practically unchanging until at the next event (build/decom/refueling) that stored state will be "uploaded."
> 
> In effect this would be cycle driven discrete dynamics and since reactors are the only ones with cycles -- reactor driven discrete dynamics. The timeline of events will be preconfigured almost entirely.
> 
> _This is most likely not worth while idea since it removes a lot of Cyclus's flexibility._
> 
> If you squint, this structure is already a dynamic of the current CEP's ideas.

The main reason I'd not go down this path is that it's too Cycamore/Reactor centric.  We don't know what other archetypes out there will implement behaviors like reactors and our solution needs to be generic.  We used to talk about being able to use Cyclus to model making toast.

gonuke:
> When you say that consumers of the commodity should always act at the same time in bullet 4, why would this be needed? Is this just another feature of that described implementation or ....
> 
> Is this intended to be a transition from
> 
> "consumer registers that they need a commodity and all suppliers (regardless of commodity output) enter to meet that request" (what we have discussed so far)
> 
> to
> 
> "trade events are registered either through request or bidding (of any commodity) and only facilities related to that commodity (suppliers or consumers) enter the DRE. The request or bid of a commodity triggers an event" Amanda had been discussing something similar for bids being necessary for events especially for back end FC.

Sort of, but let's back away from the idea that something gets registered through requests or bids.  Rather, a facility will register a desire to trade at some future instance from it's Tock.

However, when a facility registers that it wishes to trade a given commodity, we've already agreed that all suppliers of that commodity need to participate in the DRE.  I am now also suggesting that all consumers of that commodity should also participate in the DRE to ensure they don't miss out on material being available.

Consider a scenario where Facility A has registered a desire to trade Commodity X at time T, and Facility B has registered a desire to trade Commodity X at time S > T.  If only A participates in the DRE at time T, then it may be able to consume all of X such that there is none left at time S when B wishes to trade.  By forcing B to participate at time T, it can compete for the available stocks of X (if it wishes to).

meg-krieg:
> The main reason I'd not go down this path is that it's too Cycamore/Reactor centric. We don't know what other archetypes out there will implement behaviors like reactors and our solution needs to be generic. We used to talk about being able to use Cyclus to model making toast.

Makes sense -- I was just trying to toss in something different ideas.

meg-krieg:
> Sort of, but let's back away from the idea that something gets registered through requests or bids. Rather, a facility will register a desire to trade at some future instance from it's Tock.
> 
> However, when a facility registers that it wishes to trade a given commodity, we've already agreed that all suppliers of that commodity need to participate in the DRE. I am now also suggesting that all consumers of that commodity should also participate in the DRE to ensure they don't miss out on material being available.

I have questions!

1) How would a facility know when it is going have a desire to trade without looking at bids or requests? 

Suppose that some facility registers through some scheme (random or +```cycle_length``` interval or some other equation/method) in the future independent of bid/request. We know that the only outcomes of a trade are consume/supply or nothing. In order for any of those outcomes to happen doesn't the facility need to know its inventory in order to decide what sort of event/s is/are happening, and thus be going through the processes for making a bid/request anyway?  

Also in the scenario you wrote, 
> Consider a scenario where Facility A has registered a desire to trade Commodity X at time T, and Facility B has registered a desire to trade Commodity X at time S > T. If only A participates in the DRE at time T, then it may be able to consume all of X such that there is none left at time S when B wishes to trade. By forcing B to participate at time T, it can compete for the available stocks of X (if it wishes to).

What happens the event at S? Does it get unscheduled if no material is left? 

2) How would the scheduling of events be agent based if the commodity is driving the participation of agents for half of the scenarios? 

If all suppliers  and consumers of that commodity are distinctly present for at trade based on a commodity, it is the commodity driving event the happen and agents just recognizing that the commodity is trade-ready. 

3) Would this also reinforce original cyclus behavior if any time a facility has a commodity  available for trade it would trigger an event, every timestep?

gonuke:
Let me try to start wi

> > Sort of, but let's back away from the idea that something gets registered through requests or bids. Rather, a facility will register a desire to trade at some future instance from it's Tock.
> > However, when a facility registers that it wishes to trade a given commodity, we've already agreed that all suppliers of that commodity need to participate in the DRE. I am now also suggesting that all consumers of that commodity should also participate in the DRE to ensure they don't miss out on material being available.
> 
> I have questions!
> 
> 1. How would a facility know when it is going have a desire to trade without looking at bids or requests?
> 
> Suppose that some facility registers through some scheme (random or +`cycle_length` interval or some other equation/method) in the future independent of bid/request. We know that the only outcomes of a trade are consume/supply or nothing. In order for any of those outcomes to happen doesn't the facility need to know its inventory in order to decide what sort of event/s is/are happening, and thus be going through the processes for making a bid/request anyway?

I think we've maybe diverged in our vocabulary and could be saying the same thing. If I argue that every facility that does act much always act with a Tick/Trade/Tock sandwich, then once it gets to the Tock, it knows (or can know) as much as possible about its state at that time step and can make a decision about when it needs to act next.  The trade activity (bid/request) will be over at that time since we're in a Tock now, but this is the moment in a timestep of maximum knowledge on each agent's part.

> 
> Also in the scenario you wrote,
> 
> > Consider a scenario where Facility A has registered a desire to trade Commodity X at time T, and Facility B has registered a desire to trade Commodity X at time S > T. If only A participates in the DRE at time T, then it may be able to consume all of X such that there is none left at time S when B wishes to trade. By forcing B to participate at time T, it can compete for the available stocks of X (if it wishes to).
> 
> What happens the event at S? Does it get unscheduled if no material is left?

This is up to the agent that scheduled S.  In their Tock, if they are satisfied and no longer need to request an action at S, then they can unschedule it.  This does introduce an important point for the data model of my proposal - when a commodity event gets scheduled, we need to record a list of which agents may have scheduled the same event at the same time.

> 
> 2. How would the scheduling of events be agent based if the commodity is driving the participation of agents for half of the scenarios?
>
> If all suppliers and consumers of that commodity are distinctly present for at trade based on a commodity, it is the commodity driving event the happen and agents just recognizing that the commodity is trade-ready.

The original impetus has to come from an agent, but as soon as one agent wants to consume a given commodity at a given timestep, all agents who may also want to consume that commodity must be allowed to trade for that commodity, as well as all the suppliers.

Note that these agents don't need to trade for all of their commodities.

> 3. Would this also reinforce original cyclus behavior if any time a facility has a commodity  available for trade it would trigger an event, every timestep?

The whole scheme is consumer driven, so it doesn't matter who has what ready. What matters is when an agent wants to buy something.

meg-krieg:
I have an ~ extremely ~ bare bones implementation for almost every bullet *except* 4. I am still thinking about how to implement 4 and whether the registering commodity consumption should happen on the archetype side (within EnterNotify) or whether it should happen in the cyclus kernel. 

1. any agent can add itself to the execution queue for any timestamp in the future     
       - happens during Tock
2. whenever an agent acts, it must Tick/Trade/Tock this is presumably consumer-driven; there is no reason for a supplier to trigger action
3. all suppliers must be available any time that a consumer is acting they must also perform a Tick/Trade/Tock
4. all consumers of the same commodity should always act at the same time
5. Archetype developers create archetype/facility specific scheduling functions.

Would it be good to make a draft PR to Cyclus 2.0 with this implementation to check idea convergence etc/right path? And, upon any comments/new thoughts, I adjust the CEP. 

Or would it be better to adjust the CEP now for those bullet point parameters ^?

gonuke:
I'd be interested in seeing a draft implementation, to help us understand if there are nuances that need to be folded into CEP 31

